PR #7, opened 2019-05-29
Add a romstage file

rminnich:
This is all still experimental.

We need a device model, I will propose one in the next day or so.

I like the use of mod since it fixes the coreboot problem of inadvertently
calling one stage from another.


rminnich:
Apologies in advance, this is the last time I'll do this, this is really two separate things. It's early days, so I figured this one time would be ok.

rminnich:
And the invite requests are still not accepted, so in this goes.

PR #10, opened 2019-05-30
Basic PL011 driver

rjoleary:


src/drivers/uart/pl011.rs
```diff
@@ -0,0 +1,109 @@
+#![no_std]
+
+use register::register_bitfields;
+use register::mmio::{ReadOnly, ReadWrite};
+
+trait Driver {
```
rminnich: one thing I don't understand: could this driver trait be common to all our devices in oreboot? No need to do that now, just wondering.

This is cool.

rjoleary: Yeah, it definitely should be. I just didn't know which directory to put it in.

PR #14, opened 2019-06-04
WIP: ast25x0 support

rminnich:
the assembly won't assemble. I don't understand how to get the #define to work
in Rust assembly.


dancrossnyc:
#define doesn't work in Rust assembly, which is really module-level assembly as provided by LLVM. Instead, use `.set` etc.

src/mainboard/ast/ast25x0/link.ld
```diff
@@ -0,0 +1,51 @@
...
+SECTIONS
+{
+    . = 0x0;
+	.bootblock :
+	{
+	    KEEP(*(.bootblock.vectortable));
+	}
+
+    . = 0x8000;
+
```
ArthurHeymans: There is supposed to be 32K of sram at 0x1E72 0000 to 0x1E72 7FFF. It could be used as a stack to run code?

PR #19, opened 2019-06-05
Add Memory, SliceReader and Section drivers

rjoleary:


src/drivers/driver/src/lib.rs
```diff
@@ -1,46 +1,165 @@
...
+const EOF : Result<usize> = Err("EOF");
+const NOT_IMPLEMENTED : Result<usize> = Err("not implemented");
+
 pub trait Driver {
     /// Initialize the driver.
-    fn init(&self);
+    fn init(&mut self);
     /// Returns number of bytes read.
-    fn read(&self, data: &mut [u8]) -> usize;
+    fn pread(&self, data: &mut [u8], pos: usize) -> Result<usize>;
```
rminnich: so we're not doing the thing where the pread returns the amount and the "your next read would be with this offset?"

rjoleary: I found this works well for incrementing the cursor:
```
cursor += drv.pread(data, cursor)?;
```

PR #20, opened 2019-06-05
Run `cargo fmt` on crates.

dancrossnyc:


rminnich:
dan, sorry we did not move faster on this; can you fix conflicts, and rerun, and also in the PR tell us what commands you ran? This is good stuff. Thanks.

rminnich:
ah I see now, we need to add this to the README

rminnich:
we're doing this a different way and files have moved.

PR #26, opened 2019-06-20
Create a single crate for all drivers

rjoleary:
* All the drivers are in one crate.
* The crate has a hierarchy of modules.
* Some of the modules are gated by compile-time conditions.
* For example, checkout the pl011 module which is only compiled when the
  crate is imported with pl011 feature like in qemu-armv7

This also deletes the existing coreboot drivers C code.


rminnich:
so, again, how does this work across architectures? Do we really build one crate for all of x86, mips, arm, and riscv?

PR #39, opened 2019-06-27
First pass at adding a soc

rminnich:
This is missing dependencies and such so corrections welcome.


soc/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "soc"
+version = "0.1.0"
+authors = ["Ryan O'Leary <ryan@ryanoleary.ca>"]
```
rjoleary: We should probably just say "oreboot authors" in every crate.

soc/src/aspeed/ast2500/reg.rs
```diff
@@ -0,0 +1,22 @@
+const CONFIG_DRAM_ECC_SIZE: u32 = 0x10000000;
+const BASE: u32 = 0x1E60_0000;
```
rjoleary: It might be useful to use tock's register crate. See the pl011.rs for an example.

PR #48, opened 2019-07-03
Add layoutflash command

rjoleary:
The command assembles a flash image using the layout specified in a
device tree.


rjoleary:
This doesn't currently work because the boot block contains the kernel and initramfs. We need to move the kernel and initramfs into a payload in another area.

PR #91, opened 2019-08-05
add a 'log uart'

rminnich:
Just messing around. But this may be handy for debugging later
if we have jtage and can examine memory in qemu or jtag.

This brings up a questions why is print.rs appearing for each mainboard

we get very far if we don't use the sifive init()

### DIFF
--- a/src/drivers/uart/src/log.rs
+++ b/src/drivers/uart/src/log.rs
@@ -1,0 +1,48 @@
+/*
+ * This file is part of the oreboot project.
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 2 of the License.
+ *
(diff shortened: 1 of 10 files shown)


src/drivers/uart/src/log.rs
```diff
@@ -0,0 +1,48 @@
...
+impl<'a> Driver for Log<'a> {
+    fn init(&mut self) {   }
+
+    fn pread(&self, data: &mut [u8], _offset: usize) -> Result<usize> {
+        return Ok(0);
+    }
+
+    fn pwrite(&mut self, data: &[u8], _offset: usize) -> Result<usize> {
+        for (i, &c) in data.iter().enumerate() {
+            self.dat.push(c);
```
rminnich: and,yes, when it's full, it just drops it. That's fine.

PR #93, opened 2019-08-09
add ddr regs from sifive

rminnich:
won't compile, I'm doing something dumb, this is a snapshot


src/soc/sifive/fu540/src/ddrregs.rs
```diff
@@ -0,0 +1,173 @@
+/* Copyright (c) 2018 SiFive, Inc */
+/* SPDX-License-Identifier: Apache-2.0 */
+/* SPDX-License-Identifier: GPL-2.0-or-later */
+/* See the file LICENSE for further information */
+#![no_std]
+
```
rjoleary: ddrregs and phy are two separate modules, so I think you have to import phy here with something like:

import create::phy::*;

src/soc/sifive/fu540/src/phy.rs
```diff
@@ -0,0 +1,1220 @@
+/*:u32 =Copyright (c) 2018 SiFive, Inc */
+/*:u32 =SPDX-License-Identifier: Apache-2.0 */
+/*:u32 =SPDX-License-Identifier: GPL-2.0-or-later */
+/*:u32 =See the file LICENSE for further information */
+
+const DENALI_PHY_00_DATA:u32 =0x31706542;
```
rjoleary: And make all these pub.

src/soc/sifive/fu540/src/phy.rs
```diff
@@ -0,0 +1,1220 @@
+/*:u32 =Copyright (c) 2018 SiFive, Inc */
+/*:u32 =SPDX-License-Identifier: Apache-2.0 */
+/*:u32 =SPDX-License-Identifier: GPL-2.0-or-later */
+/*:u32 =See the file LICENSE for further information */
```
rjoleary: ":u32 =" -- I think you went a bit overboard with find-and-replace in here.

rminnich: ha ha I man I was tired when I did that :-)

Issue #80, opened 2019-08-02
Use llvm objcopy instead of gcc objcopy


rjoleary:
Wouldn't it make more sense to use the llvm objcopy because we're using the llvm objcopy?

andre-richter:
For reference: https://github.com/rust-embedded/cargo-binutils

andre-richter:
Cool thing is, you can then invoke the llvm binutils using your custom Rist target.json, e.g.:

https://github.com/rust-embedded/rust-raspi3-OS-tutorials/blob/master/11_exceptions_groundwork/Makefile#L78-L79

(Not needed for objcopy though).

rjoleary:
Thanks, it works really well!

PR #97, opened 2019-08-10
Use llvm objcopy provided by cargo

rjoleary:
This makes more sense than using the GCC objcopy since Rust is already
using llvm.


rjoleary:
Fixes #80

PR #102, opened 2019-08-11
Add QEMU CI test for RISC-V

rjoleary:
This simply checks that the string TESTTESTTEST is printed. This is
fairly basic and we can setup some more sophisticated in the future.


rjoleary:
Fixes #101

PR #103, opened 2019-08-11
Perform nightly build to catch breakages

rjoleary:


rjoleary:
This is to catch it when rust nightly breaks us. I remember something broke us the other week. I'll keep an eye on how many cycles we have.

PR #115, opened 2019-08-17
Clock Init

rjoleary:


src/soc/sifive/fu540/src/clock.rs
```diff
@@ -376,59 +275,37 @@ impl Clock {
...
-        // 	// get all DDR out of reset
-        // 	write32(&prci->devicesresetreg,
-        // 		PRCI_DEVICESRESET_DDR_CTRL_RST_N(1) |
-        // 		PRCI_DEVICESRESET_DDR_AXI_RST_N(1) |
-        // 		PRCI_DEVICESRESET_DDR_AHB_RST_N(1) |
-        // 		PRCI_DEVICESRESET_DDR_PHY_RST_N(1));
-
-        // 	// HACK to get the '1 full controller clock cycle'.
+
+        // HACK to get the '1 full controller clock cycle'.
```
rminnich: I think this comment should stay in, it is required.

src/soc/sifive/fu540/src/clock.rs
```diff
@@ -376,59 +275,37 @@ impl Clock {
...
         architecture::fence();
```
rminnich: you can change it to 
// required delay to get '1 full controller clock cycle'

src/soc/sifive/fu540/src/clock.rs
```diff
@@ -376,59 +275,37 @@ impl Clock {
...
+
+        architecture::fence();
     }
 }
 
 // NASTY. We'll live with it for now.
 const UART_DEVICES: u32 = 2;
 const UART_REG_DIV: u32 = 0x18;
-const UART_DIV_VAL: u32 = 4;
+const UART_DIV_VAL: u32 = 4340;
```
rminnich: I wonder if due to other bugs coreboot never got this right ...

src/soc/sifive/fu540/src/clock.rs
```diff
@@ -444,347 +321,8 @@ fn update_peripheral_clock_dividers() {
...
-/* DDR PHY reset (active low) */
-/* 4 */
-/* Reserved */
-/* 5 */
-/* GEMGXL_RST_N */
-/* RW */
-/* 0x0 */
-/* Gigabit Ethernet Subsystem reset (active low) */
+// TODO: There might be a better way to detect whether we are running in QEMU.
+fn is_qemu() -> bool {
```
rminnich: NICE!

src/soc/sifive/fu540/src/clock.rs
rjoleary: Yes, 4 is definitely wrong. The datasheet even says the value cannot be less than 16.

PR #119, opened 2019-08-24
Pass drivers to clock init using ClockNode trait to avoid nasty hack

rjoleary:
This also adds the register block for a SPI driver.


src/mainboard/sifive/hifive/src/main.rs
```diff
@@ -6,31 +6,38 @@
...
+use spi::SiFiveSpi;
 
 #[no_mangle]
 pub extern "C" fn _start() -> ! {
-    let mut v: Vec<u8, U1024> = Vec::new();
-    let mut uarts = [&mut SiFive::new(/*soc::UART0*/ 0x10010000, 115200) as &mut dyn Driver];
-    let console = &mut DoD::new(&mut uarts[..]);
-    console.init();
-    console.pwrite(b"Welcome to oreboot\r\n", 0).unwrap();
+    let uart0 = &mut SiFive::new(/*soc::UART0*/ 0x10010000, 115200);
```
rminnich: I'm wondering about removing the DoD usage, since I do want a memory console log at some point, but I guess we can always bring it back.

rjoleary: The problem here is the DoD couldn't borrow the mutable UART driver at the same time it is borrowed by the Clock struct. I'm not sure if there's a way (and when you think about it, this might be a nice property to guarantee no one is writing to the uart as it is being flushed and the divisor is being changed). We can easily create a new DoD after the clocks are initialized.

PR #122, opened 2019-08-24
Add hifive setup documentation

rjoleary:


Documentation/sifive/setup.md
```diff
@@ -0,0 +1,57 @@
...
+2. Confirm the build works by running Oreboot in QEMU. If it works, you should
+   see the words "Welcome to oreboot" and "TESTTESTTEST" printed to your
+   terminal.
+
+```
+cd src/mainboard/sifive/hifive
+cargo make -p release run
+```
+
+| ⚠️ **WARNING:** QEMU and hardware use different linker scripts. Before proceeding to the next step, you must manually copy link_hw.ld to link.ld and "rm -rf" the target directory. |
```
rminnich: wow that's awful. Why not just build each image each time? it costs little to run ld twice. Leave on in _qemu and one in _hifive?

rjoleary: I support this idea, but I don't know enough about cargo and cargo-make to get this done quickly. The linker script is referred to from the ".cargo/config" file, which is separate from the Makefile.toml. I think there should be some way of using two separate profiles in the ".cargo/config" similar to how we currently have a debug&release profile.

I created issue #123 to followup on this. Alternatively, I'm hoping there some simple flag to enable position-independent-code and then we're free of this mess altogether.

PR #130, opened 2019-08-30
Updating DDR test

pallaud:
This should improve on the previous test by wrapping around after the values read have been checked to start writing again with a different seed value - so instead of 

read 1 from index 0
read 2 from index 1
...

it'll be

read 2 from index 0
read 3 from index 1

and this will continue for 100 iterations.


pallaud:
@rminnich is this how you were thinking of testing?

GanShun:
The test might fail, due to the timeout

PR #144, opened 2019-09-06
Add HiFive device tree from coreboot

rjoleary:


philhug:
@alistair23 the firmware currently has to provide the DT to the kernel.

As the kernel is not part of the build process of oreboot, I don't see any way to get it from there. Maybe it should? ideas?

alistair23:
When you build the Linux kernel you should get a Image and a DT. It's then up to the user to supply that information to oreboot to build the flash bin file.

Wherever the kernel comes from the DT should also come from the same place. Hard coding a DT in the firmware is only going to lead to issues.

philhug:
Agreed. It might make sense to put a known good kernel+DT combination into a separate repo though, to make testing easier.

alistair23:
Maybe, there is always the GPL issue though. You need to make sure the source code is available.

Either way, it shouldn't be in the oreboot repo. For testing users should build their own kernel or maybe just use the Fedora/Debian kernel.

philhug:
The DT is useful for oreboot itself. Instead of hardcoding addresses for SPI, UART,... this info could be loaded from the DT.

rjoleary:
> The DT is useful for oreboot itself. Instead of hardcoding addresses for SPI, UART,... this info could be loaded from the DT.

This is an interesting train of thought.

It could be a fix to the problem of gating the DDR init code on `is_qemu()`. Instead, the DT in QEMU doesn't have a node for the DDR controller, so the driver for the controller is never used. The DT used in hardware does have a node for the DDR controller, so the driver for the controller is used.

However, how would the build system know at compile time which drivers to include? It might also be adding a lot of complexity...

alistair23:
I agree that adapting oreboot on the DT could be useful (let's have that discussion elsewhere) but even if we are using a DT it should be the DT from the kernel and not a custom one. I don't see any good justifcation for having a DT in the oreboot source. At best it's just duplication and at worst it opens you up to bugs as that DT differs to the kernel/U-Boot/whatever-else DT.

rminnich:
but wait. You just mentioned the u-boot DT. u-boot carries its own DT as part of build process. Why should oreboot not?

alistair23:
Argh! U-Boot does.

But the U-Boot DT applies only to U-Boot. U-Boot uses it for runtime init work. It doesn't pass it to the kernel which is what this one is doing.

philhug:
So the assumption should be to use the builtin DT for oreboot only and the user needs to provide one for the kernel, which is only used for the kernel and not for oreboot.
Agreed?

alistair23:
I would say that the user MUST provide one for the kernel.

The question remains if we want to use one for oreboot. If we do, I think we should consider just using the kernel DT, but we could also have one in oreboot.

rminnich:
This is very unlike the PC world. In the PC world, firmware always provides tables. All of them. User provides none. How can oreboot use the kernel DT if we don't know a priori which kernel we will run? From the Sun world, where this all began, open boot always provided kernels with the table too. on arm chromebooks, coreboot/depthcharge provide a DT, I was just looking.

further, IIUC, fu540 provides a basic DT from the mask rom. Isn't that the source of truth for that info?

alistair23:
> How can oreboot use the kernel DT if we don't know a priori which kernel we will run?

That is why U-Boot has it's own DT. I'm _hoping_ that the kernel DT is stable so once it's merged upstream we can rely on it being stable.... That might not be true though.

If we can't rely on it being stable then I do see the case of having a DT for oreboot to use. That should never be passed to the kernel though.

> further, IIUC, fu540 provides a basic DT from the mask rom. Isn't that the source of truth for that info?

That DT has bit rotten to being completely useless for everything. It won't work with U-Boot or Linux. That's why you don't put a DT in your firmware before you upstream the drivers/compat strings to Linux

rminnich:
is it ok if we use this dt and NOT pass it on then? We need to have something. I'm hearing we can't trust the system below us and we should not pass it on to above, but ... don't we need this one then?

alistair23:
Why not just use the DT from the kernel you are providing?

That is expect users to build a kernel and a DT.

For CI that gets harder, maybe we could have a repo with CI images and use Git LFS to handle them?

rminnich:
what kernel? What if we're booting ! Linux? Then what? Are there things in the DT we need to boot correctly and, if so, how do we get them before we load the kernel? There's something here I don't understand.

alistair23:
Ah ok. So what if you are booing not Linux. oreboot still needs a DT that it can use, even though it won't pass it to the guest.

I guess in that case we either have to check in DTs and maintain them or we could use the DT provided to us. The Unleashed DT would work enough for oreboot. The problem is that not all boards will provide a DT.

This just seems like a hassle though to maintain DTs here. If the DT is built into oreboot then why bother having them, just use hard codes in Rust.

alistair23:
I think oreboot itself should use the DT provided by the ROM/FW before it.

If nothing is provided then oreboot will just have to use hardcoded values.

If the next stage wants a DT then that stage has to provide a DT that can be built with oreboot.

I don't see an advantage in maintaining trees here

PR #146, opened 2019-09-06
Added root makefile and tool update

sphinxc0re:


rjoleary:
> Shouldn't this commit should be squashed with the previous commit?

Yeah, that's definitely a good plan going forward

PR #147, opened 2019-09-06
Split HiFive mainboard into hardware and qemu

rjoleary:
There were so many difference between hardware and QEMU (especially with
device tree) that the is_qemu function did not suffice. We also want to
use virt instead of sifive_u machine in QEMU.


alistair23:
I haven't reviewed the code, but in general here is what I think (related to QEMU).
 - Add support to Oreboot for the QEMU virt board (this should be pretty easy, I had something running yesturday)
 - Write the HiFive board to target hardware (remove the QEMU specific stuff)
   - Aim to have the HiFive board also run on the QEMU sifive_u machine without modification and without is_qemu() checks

rjoleary:
> * Add support to Oreboot for the QEMU virt board (this should be pretty easy, I had something running yesturday)

Agreed! This will come next.

> Write the HiFive board to target hardware (remove the QEMU specific stuff)

This is done.

>   * Aim to have the HiFive board also run on the QEMU sifive_u machine without modification and without is_qemu() checks

I agree this will be useful, but tricky in some parts like DDR init which QEMU does not emulate.

alistair23:
> I agree this will be useful, but tricky in some parts like DDR init which QEMU does not emulate.

Ideally you can just do the init on QEMU where it will just be ignored.

alistair23:
Shouldn't this code be moved under emulation/sifive? Then we can also have emulation/qemu-riscv for the virt board.

As for adding emulation/sifive board I'm not sure how useful it is. Ideally the HiFive code would run on QEMU. Can this patch wait until QEMU is updated to make the sifive_u machine much more similar to hardware? The QEMU patches should be merged this week and then we can re-evaluate what the differences are between HW and QEMU. The sifive_u machine isn't well used so having a target for it might not be the best option.

rminnich:
what's up with this PR

rminnich:
This PR is needed for alistair's other PR or is it obsolete?

rjoleary:
Superseded by #168

PR #153, opened 2019-09-07
Fix warnings

jwgarber:
Currently we get blasted by pages of warnings during compilation, so this fixes all of them and turns warnings into errors to prevent them from happening again. In particular, the compiler will now insist on Rust naming conventions and flag any unused variables/imports/attributes.


rjoleary:
Thanks! This is great!

Issue #154, opened 2019-09-08
Implement panic handler to print message on serial port


philhug:
Currently panic just calls wfi on riscv without printing the error message first.

rjoleary:
Good idea! This looks really easy since the PanicInfo struct (https://doc.rust-lang.org/beta/core/panic/struct.PanicInfo.html) already implements the Display trait. So printing it should be as simple as:

```
write!(w, "PANIC: {}\r\n", info);
```

inside of https://github.com/oreboot/oreboot/blob/acd9b98c9f5eb219b5936389321d87448655ab3d/src/arch/riscv/rv64/src/lib.rs#L19

The tricky part is how do we get `w`? I think it would be okay to re-initialize the UART driver here. Under normal circumstances, it would be unsafe to have multiple instances of the UART driver, but these circumstances are special. Particularly, something has already gone wrong to end up in the panic handler and this is a desperate final scream.

There's also something to be said about keeping the dependencies of the panic handler simple. The problem with `write!(w, "PANIC: {}\r\n", info);` is it might depend on a stack. If so, you'd get no output if the stack pointer is invalid. That is probably okay (and as far as I know, there's no way to guarantee the compiler won't use the stack) and would be much better than nothing.

I'll open a quick PR to add it.

PR #157, opened 2019-09-08
hifive: Print PanicInfo in panic_handler

rjoleary:
This makes the panic_handler SOC-specific because it needs to know a
thing about UART.

Tested by placing the following in _start:

	Result::<(), _>::Err("2 + 2 = 5").unwrap();

The panic looks something like:

	PANIC: panicked at 'called `Result::unwrap()` on an `Err` value: "2 + 2 = 5"', /home/ryan/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/result.rs:1084:5

Too bad the line number is not very useful (and probably makes the
binaries non-reproducible).

The size of the bootblob increases from 31K to 36K.

### DIFF
--- a/src/arch/riscv/rv64/src/lib.rs
+++ b/src/arch/riscv/rv64/src/lib.rs
@@ -4,20 +4,12 @@
 #![feature(global_asm)]
 #![deny(warnings)]
 
-use core::panic::PanicInfo;
-
 pub fn halt() -> ! {
     loop {
         // Bug with LLVM marks empty loops as undefined behaviour.
         // See: https://github.com/rust-lang/rust/issues/28728
         unsafe { asm!("wfi" :::: "volatile") }
     }
-}
-
-/// This function is called on panic.
-#[panic_handler]
-fn panic(_info: &PanicInfo) -> ! {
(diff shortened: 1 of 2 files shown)


rjoleary:
Fixes #154

philhug:
Hardcoding the uart into the panic handler seems a bit of a hack, but it should help with debugging for now.

Some kind of global console_register/print is needed to make it easier to print from non-mainboard code.

rjoleary:
> Some kind of global console_register/print is needed to make it easier to print from non-mainboard code.

Agreed! Rust is averse to having global mutable state like this. From my current understanding, there would need to be an `unsafe` somewhere in there (which is okay if there's no other way). I'll file an issue for this.

rjoleary:
> > Some kind of global console_register/print is needed to make it easier to print from non-mainboard code.
> 
> Agreed! Rust is averse to having global mutable state like this. From my current understanding, there would need to be an `unsafe` somewhere in there (which is okay if there's no other way). I'll file an issue for this.

Issue: #159

PR #164, opened 2019-09-10
Use --prefix=/opt/riscv in documentation

rjoleary:


Documentation/sifive/setup.md
```diff
@@ -69,14 +69,11 @@ sudo apt-get install autoconf automake autotools-dev curl libmpc-dev \
 git clone https://github.com/riscv/riscv-gnu-toolchain --recursive
```
alistair23: Unrelated to this patch, why not use the upstream toolchain instead of this fork?

PR #165, opened 2019-09-11
Created build configuration tool

sphinxc0re:
I started working on a build configuration tool. Name is `chisel`.

```
[COMMENCING TOOL NAMING WAR IN 3...2...1...]
```
:grin:


rminnich:
why not cake or donut or ... I know!

hugelgupf

rminnich:
let's call it vi! 

no emacs!

rminnich:
I'm still of the opinion that we ought to have the UI be http. That separates the task at hand -- configuration -- from the way it is performed. For example, one can image programmatic interaction with this tool were it to speak http. Or Lynx, Or chrome. 

I assume this speaks some sort of graphics to some sort of library. Are you wiring in x11 here? Which widgets? one reason we all type 
make menuconfig
and not
make xconfig
is the graphical configuration for Linux quickly fell apart. There's a lesson in there.

sphinxc0re:
This first prototype will probably have a terminal based UI. I'm using *Cursive* here which uses ncurses as a backend.

rminnich:
let's make sure it works on, e.g., osx; I'm willing to test. we need to verify our entire build on osx.

rjoleary:
Please re-open if there's an update!

PR #168, opened 2019-09-18
Add support for the QEMU RISC-V Virt Machine

alistair23:
This series adds support for the QEMU RISC-V Virt machine.

This series has been tested on this QEMU branch (https://github.com/alistair23/qemu/tree/mainline/alistair/board-memory.next) which is the current latest QEMU PR with a few patches. I'll send the patches to upstream QEMU later today/tomorrow.

The goal of this series is to allow people to use oreboot on the QEMU virt machine and load all of the files via QEMU's command line options. This reduces development time for new features.


src/arch/riscv/rv64/src/bootblock.S
```diff
@@ -29,23 +29,10 @@ _boot:
...
-
-	# TODO: SOC-specific
-	# Set stack pointer into CAR (L2-LIM).
-1:	auipc t0,     %pcrel_hi(_stack_ptr)
-	lw sp, %pcrel_lo(1b)(t0)
-	#lw sp, 0(t0)
+	# Set the stack pointer
+	# 0xDEADBEEF is used to check stack overflow
+1:     auipc t0,     %pcrel_hi(_stack_ptr)
+       lwu sp, %pcrel_lo(1b)(t0)
```
rminnich: quick question: won't we need the commented out code at some point? How will we do sp setup for the harts?

src/mainboard/emulation/qemu-riscv/src/print.rs
```diff
@@ -0,0 +1,21 @@
+use core::fmt;
```
rminnich: do we need to propagate this file to every new target? can't it go $TOP/src/lib/print.rs?

src/arch/riscv/rv64/src/bootblock.S
alistair23: We probably will, but I find having commented out code like this difficult to read. There is also no information on if the code ever worked so I don't think it's very useful to keep around.

src/mainboard/emulation/qemu-riscv/src/print.rs
alistair23: Yes it can. I'll do that in the next version

src/arch/riscv/rv64/src/bootblock.S
rminnich: Ah, this code totally works, it's right from working coreboot on several RISCV platforms.
I'm ok removing it as we have coreboot for backup, but .... it works. The main hangup was gnu assembly to rust assembly conversion.

alistair23: Ah ok.

In which case when re-adding it we should take it from coreboot again. Right now it's just rotting away as no one is using it.

alistair23:
I have updated all your comments and rebased this on the current master

PR #169, opened 2019-09-19
lib/device_tree: Check if the string is UTF8

alistair23:
Instead of blindly converting any data with the first byte between 0x20
and 0xff into a string let's safely see if it is valid UTF8 string.

This fixes the strange print being seen when printing the QEMU device
tree clock-frequency. This also removes ome unsafe code!

### DIFF
--- a/src/lib/device_tree/src/lib.rs
+++ b/src/lib/device_tree/src/lib.rs
@@ -203,7 +203,10 @@ pub fn infer_type(data: &[u8]) -> Type {
     }
     if let Some(i) = data.iter().position(|&c| !is_print(c)) {
         if i == data.len() - 1 && data[i] == 0 {
-            return Type::String(unsafe { core::str::from_utf8_unchecked(&data[..data.len() - 1]) });
+            match core::str::from_utf8(&data[..data.len() - 1]) {
+                Ok(ret) => return Type::String(ret),
+                Err(_e) => {  },
+            }
         }
     }
     if data.len() == 4 {
@@ -216,5 +219,5 @@ pub fn infer_type(data: &[u8]) -> Type {
 }
 
 fn is_print(c: u8) -> bool {
-    0x20 <= c && c < 0xff
+    0x20 <= c && c < 0x7f
 }


alistair23:
The test failure here is related to the diff not applying to the latest master QEMU. I send a PR to fix the failure (https://github.com/oreboot/oreboot/pull/170).

src/lib/device_tree/src/lib.rs
```diff
@@ -203,7 +203,10 @@ pub fn infer_type(data: &[u8]) -> Type {
     }
     if let Some(i) = data.iter().position(|&c| !is_print(c)) {
         if i == data.len() - 1 && data[i] == 0 {
```
rminnich: we were wondering. Can we drop the iterator? Since there's one in the from_utf8 anyway.

if data[data.len()-1] != 0 {
return something;
}

            match core::str::from_utf8(&data[..data.len() - 1]) {
                Ok(ret) => return Type::String(ret),
                Err(_e) => {  },
            }

alistair23: Yep, updated.

src/lib/device_tree/src/lib.rs
```diff
@@ -201,9 +201,10 @@ pub fn infer_type(data: &[u8]) -> Type {
     if data.len() == 0 {
         return Type::Empty;
     }
-    if let Some(i) = data.iter().position(|&c| !is_print(c)) {
-        if i == data.len() - 1 && data[i] == 0 {
-            return Type::String(unsafe { core::str::from_utf8_unchecked(&data[..data.len() - 1]) });
+    if data[..data.len() - 1] != [0] {
```
rminnich: can you clear me up on what this means? I'm confused on the [0] notation. My mind says "vector length 1 which is 0" but that's a bit different to me than " vector with 0 as value of final element" which I think is what we were thinking.
i.e., we have a vector, final element is 0, fulfilling one constraint, but we want to make sure the rest of it is value utf-8. Does this do that?

alistair23: This is just checking the following:

is element data.len() - 1 (the last element) of data equal to 0 (we use [0] to keep Rust happy).

The actual UTF8 checking has been delegated to the from_utf8() function

alistair23: Actually, we probably don't even want this check at all.

rminnich: but just checking ... isn't
data[..data.len() - 1]
a slice, as opposed to
data[data.len() - 1]

rminnich: so it matters not about the null termination? I thought Ryan wanted a NULL at the end but I no longer see why we need it ...

src/lib/device_tree/src/lib.rs
alistair23: Nope! It looks like it is required. This does some strange value checking to ensure we don't over convert numbers to strings.

Without this I see prints like this:
>         area@8
>           description = "Empty Space"
>           offset = "\u{1}\u{0}\u{0}"
>           size = "\u{1}\u{0}\u{0}"

alistair23:
Done, I have updated this PR.

PR #170, opened 2019-09-19
azure-pipelines.yml: Use a new QEMU branch

alistair23:
The current diff does not apply to the latest master QEMU. Instead of
updating the diff let's base the testing on my branch which is being
sent to upstream QEMU.

Once the patches are in upstream QEMU we can remove all the
modifications to QEMU in the testing and documentation. For now let's
leave everything else as is and do the bare minimum to get the tests
running.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,8 +26,7 @@ steps:
   displayName: 'Build for RISC-V'
 - script: |
     set -e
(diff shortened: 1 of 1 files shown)


alistair23:
Actually, it might just be better to update the diff from the branch instead. I'll open a new PR.

PR #171, opened 2019-09-19
sifive/qemu.diff: Update for mainline QEMU

alistair23:
Upstream QEMU has updated and the old diff no longer applied. While we
are updating the diff let's just update it to include the latest patches
on the QEMU list adding support for everything oreboot needs.

We need to also update the run Makefile to use the new QEMU arguments.

Once the patches are in upstream QEMU we can remove all the
modifications to QEMU in the testing and documentation.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,7 +26,7 @@ steps:
   displayName: 'Build for RISC-V'
 - script: |
     set -e
-    git clone https://github.com/qemu/qemu && cd qemu && git checkout 46517dd4971fc1fdd5b379e72cc377626ad98160
+    git clone https://github.com/qemu/qemu && cd qemu && git checkout 084f67c9d98d520c51df24f3b355774166a03691
     git apply ../tools/soc/sifive/fu540/qemu.diff
     mkdir build-riscv64 && cd build-riscv64
     ../configure --target-list=riscv64-softmmu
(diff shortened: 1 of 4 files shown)


rminnich:
is this stuck?

alistair23:
I think so, it works on my machine.

alistair23:
Ah, there is still a diff with QEMU that is causing this breakage. I'll keep digging

azure-pipelines.yml
```diff
@@ -26,7 +26,7 @@ steps:
   displayName: 'Build for RISC-V'
 - script: |
     set -e
-    git clone https://github.com/qemu/qemu && cd qemu && git checkout 46517dd4971fc1fdd5b379e72cc377626ad98160
+    git clone https://github.com/qemu/qemu && cd qemu
```
rminnich: I'm being a bit stubborn here, but I'd still like to see a git checkout of the ref for which this patch is known to work. I don't want to get into this cycle we got into where a whole ton of PRs fail b/c our qemu patch is failing. Can you put the ref in for which this patch is known to work?

alistair23: I do see what you mean, but I don't like just hardcoding random SHAs. That leaves you open to relying on bugs in QEMU.

For the time being it's probably ok, once the next QEMU release comes out we can base it on stable releases which seems safe.

PR #173, opened 2019-09-20
Fix azure pipelines patch to qemu by checking out known-good ref of qemu

rminnich:
The messages are a bit more verbose but it is still better than
things like

error 2

which we got before

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,7 +26,7 @@ steps:
   displayName: 'Build for RISC-V'
 - script: |
     set -e
-    git clone https://github.com/qemu/qemu && cd qemu
+    git clone https://github.com/qemu/qemu && cd qemu && git checkout 46517dd4971fc1fdd5b379e72cc377626ad98160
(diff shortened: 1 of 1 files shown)


rminnich:
what on earth? hub is weird sometimes. This is the "checkout a known good ref" PR

rminnich:
I'm wondering if we shouldn't sneak this one in ahead of alistair's patch as it simple. Not sure.

alistair23:
I don't think this is a good idea, we should just update the patch for the latest QEMU.

rjoleary:
Approved to unblock PRs in the short-term. In the long-term, we should update the diff (or upstream parts of it).

rminnich:
I want that ref in there however. can you now update your PR to include the checkout of the proper ref? qemu is a moving target, so if you want to apply a patch to it, you need to pick the ref that patch works for. Else we get to where we are now where a lot of new PRs break for no reason.

rminnich:
alistair has a good PR for updating qemu. 

But we should always use a ref the patch will work against.

PR #174, opened 2019-09-25
npcm7xx skeleton

rminnich:
build fails for reasons I don't understand.


rminnich:
sorry but it's jetlag, what am I screwing up here

PR #178, opened 2019-10-04
payload: fill in load_addr, and print it before you run.

rminnich:


payloads/src/payload/payload.rs
```diff
@@ -114,12 +116,19 @@ pub struct Segment<'a> {
...
-    pub fn load(&self) {
+    pub fn load(&mut self) {
         // Copy the segments into RAM.
         for s in self.segs {
             // Copy from driver into segment.
             let mut buf = [0u8; 512];
             let mut off = 0;
+            if s.typ == stype::PAYLOAD_SEGMENT_ENTRY {
+                self.entry = s.base;
+                continue;
```
rjoleary: You don't want this continue.

rminnich: why not? entries by definition are zero length.

rminnich: ah oops did not realize we had those non zero length payload segment entry ... :-)

PR #179, opened 2019-10-07
Jump to payload's entrypoint on all harts

rjoleary:


payloads/src/external/simple/testtesttest.S
```diff
@@ -6,12 +6,43 @@
...
 _start:
+	// Spin lock
+	nop
+	nop
+	nop
+	la t0, spin_lock
+	li t1, 1
+	amoswap.w t1, t1, (t0)
+	bne t1, zero, _start
+
```
rminnich: let's pretend I take a reset right here. What happens. I think cache SRAM is not cleared on reset? I would be surprised to find it is. Doesn't this code assume it is?

rjoleary: This is a test payload and runs after oreboot. The payload runs in memory and oreboot clears its spin_lock to 0 on every boot when it is copied into memory.
Look at the _start_nonboot_hart function in src/main.rs for the spin loop which runs in oreboot.

PR #180, opened 2019-10-16
Fixes to various bugs for hifive target

rjoleary:
* Add documentation for using OpenSBI
* Allow testtesttest payload to run in S-Mode
* Fix wrong segment being used as device tree
* Reset interrupts in bootblob


payloads/src/external/simple/testtesttest.S
```diff
@@ -36,8 +36,7 @@ _start:
 
 	call check_full
 	li t1, '0'
-	csrr t2, mhartid
-	add t1, t1, t2
+	add t1, t1, a0
```
rminnich: I believe in coreboot the decision was made not to trust a0. Sure you want to do that? Fine with me if so.

src/soc/sifive/fu540/src/bootblock.S
```diff
@@ -26,6 +26,21 @@ _boot:
 	# some hardware configurations, but is always set in QEMU.
 	csrr a0, mhartid
 
+	# Reset the mtimecmp to some absurdly large number.
+	# TODO: Move to Rust
+	li t0, 0x100000000
+	li t1, 0x2004000
+	sd t0, (t1)
+	sd t0, 0x8(t1)
```
rminnich: can you put a comment here that you are setting it up for all the harts.

payloads/src/external/simple/testtesttest.S
rjoleary: This is a test payload. The problem is when the payload is run in S-mode, it cannot access mhartid CSR.

src/soc/sifive/fu540/src/bootblock.S
rjoleary: I changed it so each hart sets its own mtimecmp value.

PR #181, opened 2019-10-18
Use a smaller dtfs which fits in 16MiB flash map

rjoleary:


alistair23:
This looks just like a fitImage, have you looked at swapping?

PR #185, opened 2019-11-11
Update to QEMU 4.2

alistair23:
Update the scripts and README to use the new QEM 4.2 (without any patches).

This series also fixes the `emulation/qemu-riscv` machine and adds it to the automated testing to avoid future bit rotting.


alistair23:
Just updated without the atomic hack

alistair23:
Ping :)

PR #187, opened 2019-11-15
Add the rv32

rminnich:
There are lots of options, e.g. imc vs. imac, so we should try to keep this code clean
of all the variants. That should likely be amended by the SOC.


alistair23:
Make sure you update the tests to build RV32 targets

PR #189, opened 2019-11-15
WIP: Add OpenTitan CRB mainboard

rjoleary:


src/arch/riscv/rv32/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "architecture"
+name = "arch"
```
rminnich: are you sure this will work? It was archi at some point, we originally changed it due to a name conflict.

src/mainboard/opentitan/crb/.cargo/config
```diff
@@ -0,0 +1,7 @@
+[build]
+target = "riscv32imc-unknown-none-elf"
```
rminnich: technically, this ought to work in the cpu Cargo.toml; can you try it?

src/mainboard/opentitan/crb/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "crb"
+version = "0.1.0"
+authors = ["Oreboot Authors"]
+edition = "2018"
+build = "build.rs"
+
+[dependencies]
+arch = { path = "../../../arch/riscv/rv32"}
```
rminnich: should be soc, not arch.

src/mainboard/opentitan/crb/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+console = { path = "../../../console" }
+print = { path = "../../../lib/print" }
+model = { path = "../../../drivers/model" }
+payloads = { path = "../../../../payloads"}
+cpu = { path = "../../../cpu/lowrisc/ibex" }
```
rminnich: should be covered by the soc. 
board brings in soc
soc bring in cpu
cpu brings in arch
how we do it in coreboot and it's nice. This way, we have arch, soc, and cpu dependencies all over the place.

src/mainboard/opentitan/crb/Makefile.toml
```diff
@@ -0,0 +1,58 @@
...
+# no default tasks
+skip_core_tasks = true
+
+[env.development]
+CARGO_ARGS = "--verbose"
+TARGET_DIR = "target/riscv64imac-unknown-none-elf/debug"
+
+[env.release]
+CARGO_ARGS = "--release --verbose"
+TARGET_DIR = "target/riscv64imac-unknown-none-elf/release"
```
rminnich: riscv32imc?

src/mainboard/opentitan/crb/Makefile.toml
```diff
@@ -0,0 +1,58 @@
+[config]
+# no default tasks
+skip_core_tasks = true
+
+[env.development]
+CARGO_ARGS = "--verbose"
+TARGET_DIR = "target/riscv64imac-unknown-none-elf/debug"
```
rminnich: ?? wrong cpu type

src/mainboard/opentitan/crb/Makefile.toml
```diff
@@ -0,0 +1,58 @@
...
+
+[tasks.build]
+dependencies = [ "install-rust-src" ]
+toolchain = "nightly"
+command = "cargo"
+args = ["xbuild", "@@split(CARGO_ARGS, )"]
+
+[tasks.run]
+dependencies = ["default"]
+command = "qemu-system-riscv64"
```
rminnich: will this work for rv32?

src/mainboard/opentitan/crb/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+#![deny(warnings)]
+
+use core::intrinsics::transmute;
+use core::panic::PanicInfo;
+use core::sync::atomic::{AtomicUsize, Ordering, spin_loop_hint};
+use core::fmt::Write;
+use model::Driver;
+use payloads::payload;
+use print;
+use uart::sifive::SiFive;
```
rminnich: nope

rminnich: https://github.com/lowRISC/opentitan/blob/master/sw/device/lib/uart.c to see how it works. Wow, the opentitan sw tree is .. hmm.

src/mainboard/opentitan/crb/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+use wrappers::{Memory, SectionReader, SliceReader};
+
+global_asm!(include_str!("../../../../../src/soc/sifive/fu540/src/bootblock.S"));
```
rminnich: nope

src/mainboard/opentitan/crb/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+global_asm!(include_str!("../../../../../src/soc/sifive/fu540/src/init.S"));
+
+// All the non-boot harts spin on this lock.
+static SPIN_LOCK: AtomicUsize = AtomicUsize::new(0);
+
+#[no_mangle]
+pub extern "C" fn _start_nonboot_hart(hart_id: usize, fdt_address: usize) -> ! {
```
rminnich: nuke this, we have no nonboot harts.

src/arch/riscv/rv32/Cargo.toml
rjoleary: It seems to work. Lets see if it passes the tests.

src/mainboard/opentitan/crb/.cargo/config
rjoleary: No, it doesn't seem to work. It appears to use the .cargo at the topmost layer.

src/mainboard/opentitan/crb/Cargo.toml
rjoleary: We need to import the arch to call "arch::halt()".

src/mainboard/opentitan/crb/src/main.rs
rjoleary: Done!

src/mainboard/opentitan/crb/src/main.rs
rjoleary: Done!

src/mainboard/opentitan/crb/src/main.rs
rjoleary: This is a placeholder until we get a opentitan driver written. Is there a spec for it?

src/mainboard/opentitan/crb/Makefile.toml
rjoleary: Fixed!

src/mainboard/opentitan/crb/Makefile.toml
rjoleary: Fixed!

src/mainboard/opentitan/crb/Cargo.toml
rjoleary: Fixed! Except, we still need to import the arch to call "arch::halt()".

src/mainboard/opentitan/crb/Cargo.toml
rminnich: has to be a better way. Just has to.

src/mainboard/opentitan/crb/Cargo.toml
rminnich: The problem is that the soc might implement halt, via some soc-dependent mechanism, or the cpu, or the arch. So I would argue there should be a soc::halt, which might end up calling cpu::halt, which might just default down to arch::halt

or am I crazy

rjoleary:
The build is currently failing with error messages such as:

```
error: linking with `rust-lld` failed: exit code: 1
  |
  = note: "rust-lld" "-flavor" "gnu" "-L" "/home/ryan/repos/oreboot/src/mainboard/opentitan/crb/target/sysroot/lib/rustlib/riscv32imc-unknown-none-elf/lib" "/home/ryan/repos/oreboot/src/mainboard/opentitan/crb/target/riscv32imc-unknown-none-elf/release/deps/crb-d971ec7d0b119a72.crb.ae8f5638-cgu.3.rcgu.o" "-o" "/home/ryan/repos/oreboot/src/mainboard/opentitan/crb/target/riscv32imc-unknown-none-elf/release/deps/crb-d971ec7d0b119a72" "--gc-sections" "-L" "/home/ryan/repos/oreboot/src/mainboard/opentitan/crb/target/riscv32imc-unknown-none-elf/release/deps" "-L" "/home/ryan/repos/oreboot/src/mainboard/opentitan/crb/target/release/deps" "-L" "/home/ryan/repos/oreboot/src/mainboard/opentitan/crb/target/sysroot/lib/rustlib/riscv32imc-unknown-none-elf/lib" "-Bstatic" "/home/ryan/repos/oreboot/src/mainboard/opentitan/crb/target/sysroot/lib/rustlib/riscv32imc-unknown-none-elf/lib/libcompiler_builtins-ddf3439031b17ea4.rlib" "-Tlink.ld" "-Bdynamic"

  = note: rust-lld: error: undefined symbol: __atomic_load_1
          >>> referenced by compiler_builtins.bsqadl41-cgu.7
          >>>               compiler_builtins-ddf3439031b17ea4.compiler_builtins.bsqadl41-cgu.7.rcgu.o:(__llvm_memcpy_element_unordered_atomic_1) in archive /home/ryan/repos/oreboot/src/mainboard/opentitan/crb/target/sysroot/lib/rustlib/riscv32imc-unknown-none-elf/lib/libcompiler_builtins-ddf3439031b17ea4.rlib
          >>> referenced by compiler_builtins.bsqadl41-cgu.7
          >>>               compiler_builtins-ddf3439031b17ea4.compiler_builtins.bsqadl41-cgu.7.rcgu.o:(__llvm_memmove_element_unordered_atomic_1) in archive /home/ryan/repos/oreboot/src/mainboard/opentitan/crb/target/sysroot/lib/rustlib/riscv32imc-unknown-none-elf/lib/libcompiler_builtins-ddf3439031b17ea4.rlib
          >>> referenced by compiler_builtins.bsqadl41-cgu.7
          >>>               compiler_builtins-ddf3439031b17ea4.compiler_builtins.bsqadl41-cgu.7.rcgu.o:(__llvm_memmove_element_unordered_atomic_1) in archive /home/ryan/repos/oreboot/src/mainboard/opentitan/crb/target/sysroot/lib/rustlib/riscv32imc-unknown-none-elf/lib/libcompiler_builtins-ddf3439031b17ea4.rlib
```

With rv32imc, the compiler should not be generating atomic load instructions at all.

rminnich:
filed an issue. 

Also, on further thought, I really would like to change the call to soc::halt(), if you do not mind.

rjoleary:
Here's the bug to support riscv in heapless: https://github.com/japaric/heapless/issues/123

rminnich:
https://github.com/rust-lang/compiler-builtins/issues/326

src/drivers/uart/Cargo.toml
```diff
@@ -11,7 +11,10 @@ opt-level = 'z'  # Optimize for size.
 model = { path = "../model"}
 clock = { path = "../clock"}
 register = "0.3.2"
-heapless = "0.4.x"
+
+# Heapless does not build for rv32imc.
+[dependencies.heapless]
+optional = true
```
bjorn3: You should add `version = "0.4"`

PR #192, opened 2019-11-23
fix halt. Add tool to create rust from .json. add opentitan uart.

rminnich:


rminnich:
this builds for me, what's up with out test?

rjoleary:
> this builds for me, what's up with out test?

The test seem to pass?

alistair23:
I haven't had a chance to look through this in detail, but a tool to generate rust code from the OT json would be awesome. I was thinking on adding the same thing for Tock, hopefully instead we can just extend this one to work for both.

gkelly:
Our hjson format isn't fixed, it gets new features not-infrequently.  It would probably make sense to add rust generation to the opentitan tooling and consume that output instead of having an out-of-tree parser.

alistair23:
That is a good idea. Auto generating the Tock register interface would be super useful.

PR #193, opened 2019-12-03
azure-pipelines.yml: Update to use QEMU v4.2.0-rc3

alistair23:


alistair23:
We were previously using some random SHA, now we are using a real tag. Once 4.2 comes out we can update to that and then avoid using RCs

PR #197, opened 2019-12-06
Use github.com/oreboot/simbin for qemu

rminnich:


rjoleary:
It looks like the error is:

```
[cargo-make] INFO - Execute Command: "qemu-system-riscv64" "qemu-system-riscv64" "-m" "1g" "-machine" "sifive_u,start-in-flash=true" "-nographic" "-device" "loader,addr=0x20000000,file=target/riscv64imac-unknown-none-elf/release/oreboot.bin" "-bios" "none" "-smp" "4"
qemu-system-riscv64: -device loader,addr=0x20000000,file=target/riscv64imac-unknown-none-elf/release/oreboot.bin: 'loader' is not a valid device model name
```

alistair23:
Can we not do this?

Having binaries then makes it really hard to update. QEMU 4.2 is about to be released, why not just install QEMU using the ditros packages?

gevorgyana:
I suggest that we close this in favour of #298 and #299, as QEMU v.5.x is now released and even if we want to use a pre-compiled version of it in Azure, we probably want to use a recent version.

PR #204, opened 2019-12-11
Changes to get verilator running

rminnich:
Verilator now runs with these changes, but goes down in flames
at the first sw.

### DIFF
--- a/src/soc/opentitan/earlgrey/src/bootblock.S
+++ b/src/soc/opentitan/earlgrey/src/bootblock.S
@@ -21,6 +21,72 @@
 _boot:
 	// These codes need to be implemented on a specific SoC.
 //	.weak cache_as_ram
+	# The previous boot stage passes these variables:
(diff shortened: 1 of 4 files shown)


src/soc/opentitan/earlgrey/src/bootblock.S
```diff
@@ -21,6 +21,92 @@
...
+	li t0, 0x08000000
+	li t1, 0x08010000
+	li t1, 0x08000100
+clear_bss:
+	# verilator crashes and burns on the first sw.
+	sw zero, (t0)
+	addi t0, t0, 8
+	bne t0, t1, clear_bss
+
+	# Decide if this is the boot hart.
```
gkelly: There's only one hart in the OpenTitan design, which can simplify the startup code here.

rminnich:
This gets into the rust code running and trying to do uart output on verilator!

PR #205, opened 2019-12-13
Update hjson converter to support pinmux.hjson

rjoleary:


tools/opentitanregisters/main.go
```diff
@@ -72,43 +115,98 @@ func convert(r io.Reader) (*otdev, error) {
 	return o, nil
 }
 
+func paramLookup(o *otdev, param string) string {
+	for _, p := range o.ParamList {
+		if p.Name == param {
+			return p.Default
+		}
+	}
+	log.Fatalf("Could not find paramter %q", param)
```
rminnich: parameter

could e get a bit more context here? This is going to be very hard to find maybe in a long file. I thought I had a lineno variable but I guess not. In any case, we might add one and at least print the line we're on?

rjoleary: Fixed the naming

PR #206, opened 2019-12-13
tweak uart register

rminnich:
verilator shows the hjson is wrong.


src/drivers/uart/src/opentitan.rs
```diff
@@ -29,8 +29,12 @@ use register::register_bitfields;
 #[repr(C)]
 pub struct RegisterBlock {
     ctrl: ReadWrite<u32, CTRL::Register>, /* UART control register */
+    // tracer shows we need padding?
```
GanShun: Add TODO to use new register crate

rminnich:
Yes this is a hack. We're going to go to the newest register crate which allows us to set offsets. But not yet.

andre-richter:
Ideally we can close https://github.com/tock/tock/issues/1482 before you start with that activity.

PR #210, opened 2019-12-14
Fix the dts and have the Makefile use it; other little fixes

rminnich:
The real part I thought had 512K, but verilator only takes 64k.

Regardless, clean up the dts and use that now.

The payload loading with the slice fails now, it did not before,
help.


rminnich:
we can now use the real image with embedded dts.
I messed something up as it blows up somewhere; please look.

PR #212, opened 2019-12-19
In-place sed for debug fixed-dtfs.dts

### LINKED ISSUE #135: Fix debug mode
Currently, oreboot only builds when you include the "-p release" option. Either we should fix debug mode or remove it entirely to avoid confusion.

tnballo:
**What:**

* Fixes #135 by inserting the `sed` suggested by @rjoleary, and it’s inverse, into `Makefile.toml`.
* `s#/release/#/debug/#` instead of `s/release/debug/`, to only do the replacement if word is within a path, in case there’s intentional uses of these words in the DTFS going forward (your use of the Device Tree is…interesting :wink:).

**Why:**

To address the comment about the value of debug on #135: the default debug profile includes `overflow-checks = true`, so the target will panic on integer overflow where the developer hasn’t made it explicit (e.g. [u32::wrapping_add](https://doc.rust-lang.org/std/primitive.u32.html#method.wrapping_add)). Release mode will silently allow wrap-around to happen, like C/C++. This may make a debug target worth keeping around, maybe there are other reasons – not too familiar as I’m new to Rust.

**Caveat:**

There is a risk of creating "thrashing" in the git history for `fixed-dtfs.dts` if developer A builds debug, pushes updated `fixed-dtfs.dts`, developer B clones and builds release, pushes updated `fixed-dtfs.dts`, etc. But IMO this isn’t a major drawback and is preferable to the current state (following along with the [build instructions](https://github.com/oreboot/oreboot#building-oreboot) fails for first-time users!).

**Next:**

Update `Makefile.toml` files for other targets, as appropriate, if this solution is desirable – your call.

### DIFF
--- a/src/mainboard/sifive/hifive/Makefile.toml
+++ b/src/mainboard/sifive/hifive/Makefile.toml
@@ -5,14 +5,17 @@ skip_core_tasks = true
 [env.development]
 CARGO_ARGS = "--verbose"
 TARGET_DIR = "target/riscv64imac-unknown-none-elf/debug"
+DTS_SED_INPLACE = "s#/release/#/debug/#"
 
 [env.release]
 CARGO_ARGS = "--release --verbose"
 TARGET_DIR = "target/riscv64imac-unknown-none-elf/release"
+DTS_SED_INPLACE = "s#/debug/#/release/#"
 
 [tasks.default]
 dependencies = ["bootblob"]
 script = [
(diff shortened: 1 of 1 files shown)


rjoleary:
Thank you for fixing the debug builds and for the great write-up! Please feel free to update the other `Makefile.toml` files as well.

PR #220, opened 2020-01-20
[WIP] AST25x0 Next Generation

bluecmd:
Goals:

- Make oreboot boot Linux on an AST2500 EVB
- Make oreboot boot Linux in QEMU using the AST2500 EVB machine

I will strive towards not using the training code from u-boot verbatim as an additional "nice thing".


bluecmd:
Current state:

`cargo make run` says "Hi", enters rust code, spins a while, and then reboots.

PR #225, opened 2020-02-04
clueless attempt to implement a uart

rminnich:
what I'm hoping is we have a way to implement ref/deref with the Port

### DIFF
--- a/src/drivers/uart/src/i8250.rs
+++ b/src/drivers/uart/src/i8250.rs
@@ -1,0 +1,128 @@
+// http://www.ti.com/lit/ds/symlink/pc16550d.pdf
+use core::ops;
+use model::*;
+use x86_64::instructions::port::{Port};
+use register::mmio::{ReadOnly, ReadWrite};
+use register::{register_bitfields, Field};
+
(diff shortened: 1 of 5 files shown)


src/drivers/uart/src/i8250.rs
```diff
@@ -0,0 +1,128 @@
...
+            }
+            self.d.set(c as u8);
+        }
+        Ok(data.len())
+    }
+
+    fn shutdown(&mut self) {}
+}
+
+// TODO: bitfields
```
rjoleary: Is this TODO done?

src/drivers/uart/src/i8250.rs
```diff
@@ -0,0 +1,128 @@
+// http://www.ti.com/lit/ds/symlink/pc16550d.pdf
+use core::ops;
+use model::*;
+use x86_64::instructions::port::{Port};
```
rjoleary:
```suggestion
use x86_64::instructions::port::Port;
```

rminnich:
not needed

PR #228, opened 2020-02-05
null uart device.

rminnich:


src/drivers/uart/src/null.rs
```diff
@@ -0,0 +1,31 @@
+// null is a simple Null driver.
+// TODO: move it out of uart.
+use core::ops;
+use model::*;
+
+pub struct Null {
```
gkelly: What about a unit-like struct?

```
pub struct Null;
```

rminnich: you know more than I, can you tell me more.

rminnich:
I'll push this in (I need it) and then you can tell me how to make it better :-)

PR #229, opened 2020-02-05
payloadA handling

rminnich:
This properly loads a payload. Next step is to figure out how to use
our various libraries...


src/drivers/uart/src/null.rs
```diff
@@ -1,6 +1,7 @@
 // null is a simple Null driver.
 // TODO: move it out of uart.
 use model::*;
+use x86::io::{outb};
```
rjoleary:
```suggestion
use x86::io::outb;
```

src/drivers/uart/src/null.rs
```diff
@@ -14,6 +15,9 @@ impl Driver for Null {
     }
 
     fn pwrite(&mut self, data: &[u8], _offset: usize) -> Result<usize> {
+        for (_i, &c) in data.iter().enumerate() {
```
rjoleary:
```suggestion
        for &c in data.iter() {
```

src/mainboard/emulation/qemu-q35/src/main.rs
```diff
@@ -26,26 +33,35 @@ pub extern "C" fn _start(fdt_address: usize) -> ! {
     let kernel_segs = &[
         payload::Segment {
             typ: payload::stype::PAYLOAD_SEGMENT_ENTRY,
-            base: 0x80000000,
-            data: &mut SectionReader::new(&Memory {}, 0x20000000 + 0x400000, 6 * 1024 * 1024),
+            base: 0x1000020,
+            // wtf. data should not be mutable?
```
rjoleary: I see the problem. It looks like we made the unfortunate design decision that the entry point is at the beginning of a segment (which worked so far). That should be changed.

Issue #218, opened 2020-01-16
Cargo-make issues


rjoleary:
cargo-make is deficient in the following areas:

1. **Config variables**: Ideally, the end user can customize compile-time constants such as the baudrate. Something like, but not necessarily, kconfig is desired.
2. **Duplication**: Each mainboard has a Makefile.toml which is nearly identical. Ideally, some of the logic here can be shared across mainboards.
3. **Incremental builds**: Changes to some files do not trigger a rebuild. A few times, I have accidentally used an older build.

alistair23:
meson supports Rust, maybe worth swapping?

rminnich:
it's really hard as hell to get it to do an offline build. Yeah, maybe meson. OpenTitan uses it.

gkelly:
We've had some inflexibility issues with meson. @mcy could speak more to it, but I don't know if I'd take our use of it as an outright endorsement.

sagiegurari:
actually cargo-make has options for all 3

1. config variables - you can use 'functions' in command args and swap the values. you can pass env vars in cli and you can setup env via duckscript. there are more options.
2. duplication - makefile.toml can extend each other. you can import tasks and so on from multiple other makefiles. fact that in this repo there are duplicates, doesn't mean it is a limitation in cargo-make
3. didn't fully understand this one, but cargo-make has built in watch support which should trigger right build command per area you changed.

i wouldn't assume that specific implementation of it in this repo defines what it can or cannot do.

fotonick:
Propose to close, given that we moved to regular Makefiles.

doriath:
+1 for closing it.

PR #231, opened 2020-02-09
Replace cargo-make with a standard Makefile

rjoleary:


gkelly:
https://github.com/oreboot/oreboot/blob/master/README.md#ground-rules has language that needs an update with this.

src/mainboard/Makefile.common
```diff
@@ -0,0 +1,67 @@
+# This Makefile is included from src/mainboards/VENDOR/BOARD/Makefile
```
rminnich: This Makefile is included from src/mainboards/VENDOR/BOARD/Makefile
                                             in?
                                             by?

Can you set OREBOOT in here too? 
OREBOOT=$(magic)

rjoleary:
> https://github.com/oreboot/oreboot/blob/master/README.md#ground-rules has language that needs an update with this.

Thanks for bringing this up. I will leave it up to rminnich@ to determine if this is allowable (would require changes to ground rules). This attempts to fix issues in #218. Perhaps an entirely different build system is needed.

src/mainboard/Makefile.common
rjoleary: I don't understand. For example "src/mainboard/Makefile.common is included from src/mainboards/sifive/hifive/Makefile".

What is the purpose of the OREBOOT magic?

src/mainboard/ast/ast25x0/build.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
```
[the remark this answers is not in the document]
rjoleary: Without this, it seems cargo doesn't realize when the linker script changes and won't relink the project.

PR #233, opened 2020-02-11
Improve SiFive UART pread/pwrite

nategraff-sifive:
- Implement pread
- Add retry timeout requested by TODO


rminnich:
fantastic.

Issue #152, opened 2019-09-07
Add support for SiFive SPI Memmap greater than 16MiB


rjoleary:
This was how it was done in coreboot: https://review.coreboot.org/c/coreboot/+/33055

philhug:
Fixed by #234

PR #234, opened 2020-02-12
Implement SiFive SPI Driver

nategraff-sifive:
In response to #152, implements enough of the SPI driver to enable the full 256 MB memory space mapped by the SPI controller on the FU540. I tried to copy the `fu540_spi_setup()` and `fu540_spi_mmap()` APIs created in Coreboot to do this configuration.

This doesn't yet allow us to configure SPI0 to use its full memory space, though, because right now Oreboot executes exclusively out of the SPI0 memory space. Without an SD card boot flow or the ability to jump to a second boot stage not executing out of the boot flash, this is the most I can do right now.

I'm also including a patch to include an OpenOCD config file and a cargo-make target to program the Unleashed with OpenOCD. It's the most convenient programming method available to me, but if we don't want it in the project I can remove the commit from this PR.

The driver code was missing a copyright and license, so I added one. Clearly since I'm extending the file the copyright belongs to more than just SiFive, should I add others based on `git blame`?

### DIFF
--- a/src/mainboard/sifive/hifive/src/main.rs
+++ b/src/mainboard/sifive/hifive/src/main.rs
@@ -74,6 +74,14 @@ pub extern "C" fn _start_boot_hart(_hart_id: usize, fdt_address: usize) -> ! {
     clk.pwrite(b"on", 0).unwrap();
     uart0.pwrite(b"Done\r\n", 0).unwrap();
 
+    if !is_qemu() {
+        uart0.pwrite(b"Initializing SPI controller...", 0).unwrap();
+        spi0.init().unwrap();
+        spi0.setup(0, soc::spi::FU540SPICONFIG).unwrap();
+        spi0.mmap(soc::spi::FU540SPIMMAPCONFIG).unwrap();
+        uart0.pwrite(b"Done\r\n", 0).unwrap();
+    }
+
     let w = &mut print::WriteTo::new(uart0);
 
     write!(w, "## ROM Device Tree\r\n").unwrap();
--- a/src/mainboard/sifive/hifive/src/main.rs
+++ b/src/mainboard/sifive/hifive/src/main.rs
@@ -74,11 +74,24 @@ pub extern "C" fn _start_boot_hart(_hart_id: usize, fdt_address: usize) -> ! {
     clk.pwrite(b"on", 0).unwrap();
     uart0.pwrite(b"Done\r\n", 0).unwrap();
 
+    // QEMU doesn't emulate the SPI controller interface
     if !is_qemu() {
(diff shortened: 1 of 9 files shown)


src/mainboard/sifive/hifive/src/main.rs
```diff
@@ -74,6 +74,27 @@ pub extern "C" fn _start_boot_hart(_hart_id: usize, fdt_address: usize) -> ! {
     clk.pwrite(b"on", 0).unwrap();
     uart0.pwrite(b"Done\r\n", 0).unwrap();
 
+    // QEMU doesn't emulate the SPI controller interface
+    if !is_qemu() {
```
rminnich: my only slight concern here is shipping binaries in boards with this test in. Is there some way we can avoid that or is there magic I don't understand that ensures we avoid it?

nategraff-sifive: The `is_qemu()` test or the read from 0x3000_0000? There's another `is_qemu()` at line 105 so I figured that was okay. I can take out the read from 0x3000_0000.

andre-richter: Regarding this is_qemu() check: In x86 world checking for QEMU is often done by inspecting certain flags in the CPUID return values (don’t remember which right now).

I know nothing about RISCV or how QEMU emulates it, maybe something like that is possible as well, where you don’t read from a memory location that might have unforeseen consequences on real hardware?

nategraff-sifive: [Right now Oreboot reads the MSEL register](https://github.com/oreboot/oreboot/blob/master/src/soc/sifive/fu540/src/lib.rs). Apparently QEMU sets that register to 0x297, which isn't possible on actual hardware. I can't find where QEMU does that, but it does seem to work. I'll also note that there's a big ol' TODO right next to the check that says there might be a better way.

andre-richter: I have to admit that I did not see that the is_qemu test was local for the respective board that is emulated here. I thought this was a universal test for anything riscv. Apologies for the confusion.

Just in case a board-agnostic test is wanted sometime in the future:
I just took a quick browse through the ISA. Is QEMU setting something unique in the mimpid cpu Register that maybe differentiates between QEMU and real HW?

nategraff-sifive: Unfortunately, the mvendorid/marchid/mimpid weren't implemented at the time that the FU540 SoC was created, so the hardware returns zeros. I suppose that means QEMU could return nonzero values, the trick is which ones?

rminnich: I think this is OK for now, just wanted to get it out there.

PR #237, opened 2020-04-02
Reduce size of OpenTitan CRB ROM image

rjoleary:
This redues the size from 16KB to 7.4KB.


rjoleary:
There is an explanation here on why this reduces code size: https://jamesmunns.com/blog/fmt-unreasonably-expensive/

PR #239, opened 2020-04-10
WIP: add ioport abstraction

rminnich:
won't build for e.g. sifive/hifive, help accepted.

We need this so a uart can use an ioport across, e.g., x86 and arm.

Further, memmory mapped IO ports may come with restrictions that memory does not,
e.g. around size or alignment, so an IO port abstraction, even for memory, is helpful.


jackpot51:
These should be volatile reads and writes

rminnich:
on discussion, we realized this is not quite what we want.

src/arch/arm/armv7/src/lib.rs
```diff
@@ -48,6 +49,27 @@ impl Driver for MMU {
     fn shutdown(&mut self) {}
 }
 
+/// The driver reads directly from memory, but there may be restrictions
+/// around word alignment and operation size.
+pub struct IOPort;
+
+impl Driver for IOPort {
+    fn pread(&self, data: &mut [u8], pos: usize) -> Result<usize> {
+        let src = unsafe { from_raw_parts(pos as *const u8, data.len()) };
```
sphinxc0re, line 58: Do you have to use `unsafe` here? You could just create another slice:
```rust
let src = &data[pos..];
```

sphinxc0re: oh... too late

Issue #242, opened 2020-04-24
Migrate asm! macro, which is no longer supported on nightly


fotonick:

fotonick:
It seems that `asm!` is currently broken on nightly rustc and we're being prodded toward `llvm_asm!` as a stopgap until they finalize a new `asm!` macro. We need to adapt.

[An alternate solution](https://stackoverflow.com/questions/15354488/how-to-embed-llvm-assembly-or-intrinsics-in-c-program-with-clang), which @rjoleary and @rminnich seemed to like at first listen, is to replace our inline ASM with some LLVM IR and mark `alwaysinline`. No more unstable nightly macro. (Still have to mark call sites with `unsafe`, of course.)

fotonick:
Note that #244 got the build working using the `asm` → `llvm_asm` replacement. I'll close this issue for now, but keep in mind the alternate solution discussed above as a possible future improvement.

PR #244, opened 2020-04-25
Fix build with latest nightly rustc and cargo objcopy

fotonick:
The current invocation seems no longer supported in cargo-binutils
0.2.0. The change herein appears to be forwards (tested 0.2.0) and
backwards (tested on 0.1.6) compatible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,9 @@ firsttime:
 	curl https://sh.rustup.rs -sSf | sh
 	rustup override set nightly
 	rustup component add rust-src llvm-tools-preview
+	rustup target add riscv64imac-unknown-none-elf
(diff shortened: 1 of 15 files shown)


fotonick:
In the end, I had to fix previous issues to get the Azure build going. I found that we were missing toolchains for some of the target architectures, so updated the Makefile. Updating the Makefile wasn't enough, as the Azure CI had a totally independent, nearly-identical command to set up, so I made a choice to make the firsttime target completely non-interactive then have CI call `make firsttime`. I'm back to a building state with the latest ``cargo objcopy`` and latest nightly compiler.

fotonick:
Note that this PR resolves issues #242 and #245.

PR #248, opened 2020-04-27
Use rust-objcopy instead of cargo-objcopy

rjoleary:
cargo-binutils v0.2.0 now tries to rebuild your crates which was
problematic when also using cargo-xbuild.

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -42,7 +42,7 @@ $(IMAGE): $(BOOTBLOB) $(FIXED_DTFS)
 	@printf "**\n** Output: $@\n**\n"
 
 $(TARGET_DIR)/bootblob.bin: $(ELF)
-	RUST_TARGET_PATH=$(TOP)/src/custom_targets cargo objcopy --bin $(notdir $<) -- -O binary -R .bss $@
+	rust-objcopy -O binary -R .bss $< $@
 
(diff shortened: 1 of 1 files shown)


fotonick:
I believe that your solution is not correct. It looks like `rust-objcopy` is the same as `cargo-objcopy` ([ref](https://github.com/rust-embedded/cargo-binutils/tree/master/src/bin)) and it works for you only because you a) have `cargo-binutils` already installed and b) have the old version (0.1.6).

rjoleary:
Thank you Nick!

I'll try adding cargo-binutils back in.

Using rust-objcopy was based off the advice in: https://github.com/rust-embedded/cargo-binutils
It seems cargo-binutils made a breaking change to their CLI.

We really need a hermetic build system where we can pin the versions of all dependencies.

fotonick:
You're quite right that they explicitly recommend the `rust-` variants of the tools, so they're unlikely to go away.

Yeah, our Cargo.toml-external dependency management is weak right now. There are [tricks](https://www.debian.org/doc/manuals/debian-reference/ch02.en.html#_recording_and_copying_system_configuration) to record and install Debian package versions. I don't see a great way of doing that for rustup. It's a shame, since cargo is so good at this.

PR #249, opened 2020-04-27
Add x86 build to CI

rjoleary:


fotonick:
Once this works, it seems like a good time to update the top of the README, where it says no binary blobs, so no x86.

PR #255, opened 2020-05-02
CI parallelization

fotonick:
1. Do a shallow git clone to save 10 seconds. This will represent more savings as the repo gets bigger.
2. Add proper dependencies between steps to gain some parallelism.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,6 +13,8 @@ pool:
   vmImage: 'ubuntu-latest'
 
 steps:
+- checkout: self
+  fetchDepth: 1  # Download only latest commit, not all history
 - script: |
     git submodule update --init
   displayName: 'Checkout submodules'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,58 +12,97 @@ schedules:
 pool:
   vmImage: 'ubuntu-latest'
 
-steps:
-- checkout: self
-  fetchDepth: 1  # Download only latest commit, not all history
-- script: |
-    git submodule update --init
-  displayName: 'Checkout submodules'
-- script: |
-    make ciprepare
-  displayName: 'Install Rust Dependencies'
-- script: |
-    make --keep-going format check=true
-  displayName: 'Check formatting'
-- script: |
-    make --keep-going test
-  displayName: 'Run rust tests'
-- script: |
-    make --keep-going clippy
-  displayName: 'Run clippy linter'
-- script: |
-    make --keep-going mainboards
-  displayName: 'Build all mainboards'
-- script: |
-    ./scripts/generate-size-report.sh
-  displayName: 'Generate report of binary sizes'
-- script: |
-    cd src/mainboard/sifive/hifive
-    PAYLOAD_A=../../../../payloads/src/external/simple/testtesttest make
-  displayName: 'Build SiFive board for RISC-V'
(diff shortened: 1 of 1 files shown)


rjoleary:
Woah! This looks really cool!

fotonick:
Unfortunately each of the myriad approaches I tried had some issues. The crux: Each Azure Pipeline "job" is a completely fresh VM instance with nothing carried over. And unfortunately a job is the smallest unit of work for which you can request parallelism.

To actually get parallel Build jobs, I think we would need to download our own Docker image from external blob storage into the instance they give us. This Docker image would have all the dependencies preinstalled. Each build step would then be "download the Docker image, run it, clone the oreboot repo, start building". QEMU tests would run on the same instance as their corresponding builds.

fotonick:
Oh, and here's a minimal Docker file that can successfully build Oreboot (no QEMU yet):

```
FROM alpine:3.7

RUN \
  cd && \
  apk update && \
  apk add git curl gcc libc-dev make dtc pkgconf openssl-dev && \
  curl https://sh.rustup.rs -sSf | sh -s -- -y --profile minimal --default-toolchain nightly && \
  source ~/.cargo/env && \
  rustup override set nightly && \
  rustup component add rust-src llvm-tools-preview && \
  rustup target add riscv64imac-unknown-none-elf && \
  rustup target add riscv32imc-unknown-none-elf && \
  rustup target add armv7r-none-eabi && \
  cargo install cargo-xbuild cargo-binutils
ENTRYPOINT sh --login
```

Unfortunately, it's about 1GB. At this size, an Ubuntu base might be better just so that we can reuse `make firsttime`.

fotonick:
Oh, maybe it's simpler than that. Azure offers artifact caching, which lets you save and load build products between jobs. https://docs.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops

fotonick:
This patch now gives some limited parallel speedup.

- Before: There have been 19 builds of the current HEAD (`9cded18`), and they have taken anywhere from a min of 14m15s to a max of 16m30s, with a median of 15m31s.
- After: With the parallel speedup, the Azure pipeline passed in 12m5s. That's a 22% speedup relative to the median of the Before range.

I get the speedup by spawning three jobs in parallel, each of which is a separate VM instance, but they're doing highly redundant work, rerunning `make firsttime`, rebuilding common deps, rebuilding QEMU, etc. I tried publishing/download intermediate build artifacts to reduce the dependencies, but compile time was surprisingly comparable to the sum of publish and download times.

Similarly, any Docker image we build for compiling rust will be O(1GB) which costs quite a lot of time to download and didn't save much, if any, time in my early testing (I didn't record numbers, unfortunately). We may want to do the Dockerfiles anyway, just to lock versions of *everything*, but it won't be for CI speedup. 

One speedup I see is to terminate the QEMU instances based on something other than a hard-coded timer.

A long-term advantage of this approach is that we pay no more time for adding more boards or running tests on our tools. The time taken is set by the longest-running build-test.

Conclusion: I propose to merge for the 22% speedup and future expandability, even if it's slightly wasteful of compute.

orangecms:
There are conflicts now, sorry... Can you resolve them? Then I'd just pull this patch in. Nothing is more annoying than waiting too long for CI these days. 😅

rminnich (approved):

PR #263, opened 2020-05-22
Vendorcode

rminnich:


fotonick:
Can you document here why you're removing the code you pulled in just days before?

rjoleary:
> Can you document here why you're removing the code you pulled in just days before?

We're going to import exactly what is needed and laid out into the file structure for a rust crate. Just header files and blobs. Also, the FSP license is a bit peculiar for vendoring.

PR #243, opened 2020-04-24
Check Rust code formatting in CI

rjoleary:
Signed-off-by: Ryan O'Leary <ryanoleary@google.com>

PR #267, opened 2020-07-28
Add format target to Makefile.

doriath:
This new target will automatically run rustfmt on all rust code in this
repository (note: only rust code that is inside some cargo crate will be
formatted).

It also introduces special `check` parameter that we can later use in CI
to verify that all pull requests are properly formatted, by running:

```
$ make --keep-going format check=true
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ help:
 	@echo 'Make options:'
 	@echo 'firsttime -- for the first time you run make'
 	@echo 'update -- to update the install'
+	@echo 'format -- to format all files'
 	@echo '  # Build a single board'
 	@echo '  make VENDOR/BOARD'
 	@echo '  # This is equivalent to'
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -43,5 +43,18 @@ firsttime:
...
 	rustup update
 
+# Option used for formatting. If set, the command will only verify if
+# formatting is correct (without actually changing the formatting).
+# Returns 0 only if all files are properly formatted.
+# Usage:
+# 	$ make format -k check=true
+check ?=
+
+CRATES_TO_FORMAT := $(shell find . -name "Cargo.toml")
```
rminnich, line 53: I'd really like to avoid finds like this if possible. we're starting down the same messy path that the coreboot makefiles followed. The struct of the oreboot directory tree is highly constrained (intentionally); is there some way you can just glob, e.g. src/*/*/*/Cargo.toml

I've been struggling with coreboot make complexity a lot the last few months. I know this does not seem like much but in a few years time, if we are not careful, we're going to have the same makefile mess coreboot has today.

consider this: the goal of oreboot is to allow someone who is not a fulltime firmware person to build and use firmware. it's essential that we keep things simple. Is there some way to make this a bit less so.

doriath: Makefile does not support recursive "wildcard", but if we assume that max depth of any Cargo crate would be e.g. 4, then we can just call wildcard 4 times. It is a little longer (and depends on max depth), but is probably much easier to understand (does not depend on understanding custom commands like find).

On similar note, if you do not like the "check" parameter, and you would prefer to just use additional target (e.g. `check_format` or `check.format` or `ci.format`), please let me know and I can try to update it as well (but this might not be as trivial as it seems, and in the end could complicate makefile more than the variable). But if you want I can at least try it out and we can choose which solution is preferred.

doriath:
I see now that #243 is doing the same thing (adds fmt to make file and even updates the CI). Do you have any preferences for the approach?

doriath:
Could I ask for reapproval (I needed to rebase this PR)?

PR #268, opened 2020-07-30
Add comments to device tree crate.

doriath:

### DIFF
--- a/src/lib/device_tree/src/lib.rs
+++ b/src/lib/device_tree/src/lib.rs
@@ -1,3 +1,20 @@
+//! `device_tree` is a library for operating on the Device Tree data structure.
+//!
+//! Device tree is a data structure for describing the hardware. Useful abbreviations:
+//!
+//!   * FDT - Flattened Device Tree format (also called Device Tree Blob (DTB) format) is a compact
+//!           binary encoding of the device tree structure.
+//!   * DTB - Device Tree Blob format (also called Flattened Device Tree (FDT) format) is a compact
+//!           binary encoding of the device tree structure.
+//!   * DTS - Device Tree Syntax is a human friendly text representation of the device tree.
+//!   * DTC - Device Tree Compiler is a tool that can convert device tree between different
+//!           formats.
+//!
+//! The code in this library is heapless (no heap allocated data structure is used) so that it can
+//! be used in bootloader.
+//!
+//! For more information about Device Tree Standard, see https://www.devicetree.org/.
+
 #![no_std]
 #![deny(warnings)]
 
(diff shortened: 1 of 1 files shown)


doriath:
Note to reviewers: Those comments are based on my current understanding of the code, but it is possible I misunderstood something so please point out when I am incorrect.

src/lib/device_tree/src/lib.rs
```diff
@@ -9,11 +26,25 @@ use core::fmt;
 use model::{Driver, Result};
 use wrappers::SectionReader;
 
+/// Special value that every device tree in FDT format starts with.
 pub const MAGIC: u32 = 0xd00dfeed;
+// TODO: Explain why max depth is 16.
```
rjoleary, line 31: This is just an arbitrary number. The largest nesting level of device tree supported. With a fixed max, we can guarantee the stack won't overflow.

src/lib/device_tree/src/lib.rs
```diff
@@ -148,6 +192,7 @@ impl<'a> Iterator for FdtIterator<'a> {
                     self.len_buf[self.depth] = cursor_string(&self.dt.struct_block, &mut self.cursor, &mut self.path_buf[self.depth]).unwrap() as usize;
                     self.cursor = align4(self.cursor);
                     self.depth += 1;
+                    // Note: This performs a copy of the whole path_buf array!
```
rjoleary, line 195: Good eye! Ideally, we would return a reference to `self.path_buf` to avoid the copy.

The problem is the function signature is:

```
fn next(&mut self) -> Option<Entry<'a>> {
```

and not:

```
fn next(&'a mut self) -> Option<Entry<'a>> {
```

Without the extra `'a`, the the Rust borrow checker will complain about the lifetimes if we returned a reference to `self`.
We cannot chain the signature because we must implement `Iterator`.

Many people have told me we can solve this problem using "streaming iterators".

src/lib/device_tree/src/lib.rs
```diff
@@ -9,11 +26,25 @@ use core::fmt;
...
+///
+/// * the `node-name` part is required to be between 1-31 characters (bytes)
+/// * `unit-address` is optional, but if present, should contain address in hex format.
+///   We are reserving 32 characters to encode 128 bit address.
+/// * We are reserving 1 byte for null terminator.
+///
+/// An error is returned for longer names.
+///
+/// TODO: Reserve 1 byte for "@" character.
+/// TODO: Do we need to encode 128 bit addresses, or would it be enough to just support
```
rjoleary, line 45: Current the representation of the name is just a string, we don't try to parse the components (yet).

src/lib/device_tree/src/lib.rs
doriath: Thanks for explaining! I updated the comment.

src/lib/device_tree/src/lib.rs
doriath: Yes, I started looking at it and indeed it is not a trivial fix. I have some ideas I want to try, but before I start recommending bigger changes I wanted to familiarize myself with the codebase first.

Just as FYI what I want to try (outside of this PR):

- First is something similar to streaming iterators you mentioned. I am not sure if there is any code that uses this FdtIterator and takes some big advantage of the fact that this is a standard iterator and not some custom set of methods. If not, maybe we can just design better API for our usecase. With the new API, I would like to ensure that errors are also properly handled in it (as mentioned in one of the TODOs in this file, it's unclear what to do with errors in the iterators). I need a little more time and experimentation to propose something more concrete here.

- The other part I want to explore is whether we actually need the Path to contain the whole path, or could we just return the last component of the path (node name or property name), and let the user of the library properly handle reconstructing the whole path if they need to. This would have few interesting effects:

  - The copy would not be such a big problem, as we would only copy the buffer for one name (64 bytes now), instead of the whole 16*64 array of arrays.
  - This might remove the restriction on maximum nesting of the library (as we would never store the whole path, just the node we are currently in)
  - The disadvantage though is that if someone needs all components of the path, then such API might be harder to use. That is why I need to explore a little more how this library is used before proposing this change.

src/lib/device_tree/src/lib.rs
doriath: Yes, in this comment I am just trying to reverse engineer why this constant is "31 + 1 + 32" and I wasn't sure where the "32" part came from. My best bet was that it was used to reserve a little more space for this "unit-address" component. From my understanding, it is supposed to be a hex number, so 32 bytes would allow us to store the 128 bit address. This TODO is to figure out if this 32 bytes was intentional to represent 128 bits, or was it used mostly because it nicely double the other 32 bytes, and the resulting 64 will probably align well in memory.

doriath:
Thank you for review and explaining to me the code and possible improvements!

src/lib/device_tree/src/lib.rs
rjoleary: The main usecase is iterating over the FDT and pulling out some specific values and storing them in variables on the stack.  It's for configuration at runtime. The tricky part is there is only stack memory, so no heap memory.

Feel free to re-design this. You can see the examples of how the API is used right now.

Also, take a look at how libfdt is designed for some ideas: https://github.com/dgibson/dtc/blob/master/libfdt/libfdt.h#L180

src/lib/device_tree/src/lib.rs
rjoleary: Your analysis looks right. The number is basically arbitrary, but has to be "big enough". If I recall, the spec does not say anything about a maximum length. But, for practical reasons, we limit it.

As for the "31 + 1 + 32", it does look like it is intended to reserve space for different parts. Feel free to replace this with "64" if it will make the docs easier.

Issue #167, opened 2019-09-16
Improve usability of layoutflash tool


rjoleary:
* "No such file or directory"  error message should also print the file which could not be found.
* If a file is too big to fit into the flash area, an error message should be printed. Currently, the tool silently does something bizarre.
* If two or more areas overlap, print an appropriate error message.
* Automatically calculate the offset of the area based on the sizes of the previous areas.

doriath:
(1) seems already done, and #269 should resolve (2) and (3).

I will try doing (4) as well, but it will probably require adding some empty areas in some dts files (e.g. qemu-q35 has some gaps between areas).

PR #269, opened 2020-08-02
Add more error checking to layoutflash tool.

doriath:
It resolves two improvements mentioned in #167.

### DIFF
--- a/tools/layoutflash/src/main.rs
+++ b/tools/layoutflash/src/main.rs
@@ -67,9 +67,17 @@ fn read_fixed_fdt(path: &Path) -> io::Result<Vec<Area>> {
     Ok(areas)
 }
 
-fn layout_flash(path: &Path, areas: &[Area]) -> io::Result<()> {
+// This method assumes that areas are sorted by offset.
+fn layout_flash(path: &Path, areas: &mut [Area]) -> io::Result<()> {
+    areas.sort_unstable_by_key(|a| a.offset);
     let mut f = fs::File::create(path)?;
+    let mut last_area_end = 0;
     for a in areas {
+        if a.offset < last_area_end {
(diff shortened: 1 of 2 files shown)


doriath:
I missed the fact that fixed device tree in qemu-q35 had overlapping areas and areas were not properly sorted (this failed on azure pipelines). I tried to fix the file, but I had to remove two areas (`RomPayload DTFS A and B`), because they were overlapping with `RamPayload DTFS A`, but I am not sure if that is correct. Please let me know if there is a better fix for this.

tools/layoutflash/src/main.rs
```diff
@@ -67,9 +67,16 @@ fn read_fixed_fdt(path: &Path) -> io::Result<Vec<Area>> {
     Ok(areas)
 }
 
+// This method assumes that areas are sorted by offset.
 fn layout_flash(path: &Path, areas: &[Area]) -> io::Result<()> {
     let mut f = fs::File::create(path)?;
+    let mut last_area_end = 0;
     for a in areas {
+        if a.offset < last_area_end {
```
rminnich, line 75: Why do they need to be sorted by offset? Why place that burden on the user? I can tell you it gets old fast. 

Option 1: read in the file, sort it, then do this check.
Option 2: provide a tool that reads the file in and spits it out sorted (I kind of like that one)

Please don't make people hand sort files. It's a pain.

doriath: I am sorry about it, added sorting. I was thinking about adding sorting, and I did not do it initially because:

* it is slower (I know it is silly reason, especially that areas has only few elements)
* I was kind of preparing for trying to tackle last point in #167, where I believe the intent is to drop the offset and just auto-compute it based on size, and for that areas have to be sorted in file. But I should not have made this decision based on some future plan (which I have a little mixed feelings about and I am not sure if it will even be approved). 

I think it is a right decision to do the sorting here now and we can always drop it in the future if it will not be needed.

PR #270, opened 2020-08-03
Several patches to get a rome crb working.

rminnich:
gets all the way into start() in rust.


src/mainboard/amd/romecrb/fixed-dtfs.dts
```diff
@@ -0,0 +1,72 @@
...
+    flash-info {
+        compatible = "ore-flashinfo";
+        board-name = "QEMU q35";
+        category = "emulation";
+        board-url = "http://wiki.qemu.org/Main_Page";
+        areas {
+            area@0 {
+                description = "Boot Blob";
+                compatible = "ore-bootblob";
+                offset = <0xff0000>;
```
doriath: Would it be possible to sort those areas by offset and make sure there are no overlapping areas? Otherwise this will conflict with #269

rminnich: The look sorted to me, just reverse order?

PR #273, opened 2020-08-13
Extract method reading device tree header.

doriath:
It also changes the logic to first read the whole header into 40 byte
array and then convert them to integers. This has the advantage that we
check the Result from pread just once (instead 10 times), and we can
also check the actual size returned by pread (before the code did not
check that we actually read 4 bytes).

### DIFF
--- a/src/lib/device_tree/src/lib.rs
+++ b/src/lib/device_tree/src/lib.rs
@@ -115,12 +115,6 @@ pub struct FdtReader<'a> {
     strings_block: SectionReader<'a>,
 }
 
-fn read_u32(drv: &dyn Driver, offset: usize) -> Result<u32> {
-    let mut data = [0; 4];
-    drv.pread(&mut data, offset)?;
-    Ok(BigEndian::read_u32(&data))
-}
-
 fn cursor_u32(drv: &dyn Driver, cursor: &mut usize) -> Result<u32> {
     let mut data = [0; 4];
     *cursor += drv.pread(&mut data, *cursor)?;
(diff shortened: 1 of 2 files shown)


src/lib/device_tree/src/lib.rs
```diff
@@ -142,24 +136,44 @@ fn align4(x: usize) -> usize {
...
 
+// Reads the header of the device tree. 
+fn read_fdt_header(drv: &dyn Driver) -> Result<FdtHeader> {
+    // 10 fields, each field 4 bytes.
+    const HEADER_SIZE: usize = 10 * 4;
+
+    let mut data = [0; HEADER_SIZE];
+    let size = drv.pread(&mut data, 0)?;
+    if size != HEADER_SIZE {
+        return Err("not enough data to read device tree header");
```
rminnich, line 147: if would be nice at this point to tell us just how much data you were able to read.

src/lib/device_tree/src/lib.rs
```diff
@@ -142,24 +136,44 @@ fn align4(x: usize) -> usize {
...
+    }
+    let mut cursor = 0;
+    let mut read_u32 = || {
+        cursor += 4;
+        BigEndian::read_u32(&data[(cursor - 4)..cursor])
+    };
+
+    let header = FdtHeader {
```
rminnich, line 155: ah this is nice. I just learned something.

src/lib/device_tree/src/lib.rs
doriath: I would love that as well, but because the error is of type `&'static str`, I believe we can only use string literals (there is no way I know of to add custom information like number of bytes here). :(

rjoleary (approved):
LGTM!

PR #274, opened 2020-08-15
Add tests for slice reader.

doriath:
Also document Driver.pread behavior when no more bytes can be read.


src/drivers/model/src/lib.rs
```diff
@@ -12,6 +12,9 @@ pub trait Driver {
         Ok(())
     }
     /// Positional read. Returns number of bytes read.
+    ///
+    /// If there is no more bytes to read, returns `EOF` error (note: 
+    /// `std::io::Read` trait would return 0 in that situation).
```
rminnich, line 17: weird historical note: 

The particular problem of read and EOF has been discussed for decades. EOF really ought to mean "not only did this fail, but it will never succeed no matter how many times you try". 0 ought to mean "nothing there now, but you can come back later". 

The unix 'returns 0' in *C* is because C return values are so limited. In the original unix assembly API there were actually 3 return values: the carry bit,  register r0, and the the return PC (you read that right). So the option of "error, never again, try again later" -- 3 possible return values at least -- was possible in assembly in ways that it was not in C, and so here we are.

it's not clear to me that std::io:Read did any better than C. 

We may have got this distinction wrong in oreboot, but I'm not sure.

src/drivers/wrappers/tests/tests.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+#[test]
+#[should_panic]
+fn test_slice_reader_returns_eof() {
+    let data = [1, 2, 3, 4, 5];
+    let driver = SliceReader::new(&data);
+    let mut buf = [0; 5];
+    
+    driver.pread(&mut buf, 5).unwrap();
```
rminnich, line 41: fill me in -- why does this panic?

src/drivers/model/src/lib.rs
doriath: The main reason to add this comment here was to just document the API. The method is very, very similar to `std::io::Read`, and I was pretty surprised when I saw that the actual implementations return `EOF`.

I personally do not have very strong opinion either way and I also find it strange that the `std::io::Read` that does not distinguish between read 0 bytes, but will never read any more vs read 0 bytes but later maybe will return some (it seems there is no way to check if the stream is actually finished or not).

On the other note, I found few uses of `Driver` trait in this repo that seems to assume that the input buf will always be filled in. For example, if I want to read 4 bytes, often we just pass 4 element array and assume that driver will fill in all 4 bytes in one go, without taking into account that it could first read just 3 bytes, and then in additional read would return the remaining one byte (I am missing some kind of `read_all` method that would always ensure the passed in buffer is always filled in). But I need to look into more implementation of `Driver` to see if I should update the documentation of `Driver`, adding an assumption that the buffer will always be fully filled in, or if I should add some kind of `read_all` method and use it in more places.

src/drivers/wrappers/tests/tests.rs
doriath: `data` has 5 bytes, and if I try to read at pos 5, pread should return EOF. I am calling `unwrap` here, so it should panic.

PR #278, opened 2020-09-16
Use wildcard to define list of mainboards.

doriath:
The problem was the ordering of arguments inside filter-out function (in
the comment the arguments were in invalid order which is why it was not
working).

I am adding emulation/qemu-armv7 and nuvoton/npcm7xx to the list of
broken targets because they fail to build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,15 +14,12 @@ help:
 	@echo '  # Build debug mode'
 	@echo '  MODE=debug make mainboards'
 
-BROKEN := src/mainboard/ast/ast25x0/Makefile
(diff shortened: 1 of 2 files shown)


orangecms:
Legit 👍 

And filtering makes more sense than maintaining two lists, yea. Is the filter function portable or a GNUism? Otherwise I'd suggest to only add the Rome CRB to the broken-list.

doriath:
The filter-out function is part of Make so it should be portable.

fotonick (approved):

orangecms (approved):

Issue #38, opened 2019-06-24
We are overusing trait objects


rminnich:
We need to rethink our use of trait objects

sphinxc0re [1:21 AM]
I see that there are a lot of trait objects in the current codebase.

I don't think that's good. We should use generics whenever we can.
For lists of abstract objects, there is no other way though.
We couldn't do `Vec<T> where T: Driver`
because that would mean we had a list of all the same driver

rminnich [8:10 AM]
if we  used &Driver more could we then use generics instead of traits? I did get to the writeups on why generics are frequently better than traits and your point is well taken, esp. in the romstage. How can we reduce trait usage?

sphinxc0re [8:36 AM]
We would still use traits but not necessarily _trait objects_ anymore.
We can look for places in the code, where trait objects are used in a field or parameter situation. In those cases you can usually swap them out for generics that _implement that trait_
or in some rare cases just do `impl Trait`

orangecms:
This is now obsolete. Closing thusly. :-)

PR #281, opened 2020-09-17
Use generics instead of trait objects in FdtReader

doriath:
ref #38

Generics allow us to avoid dynamic dispatch and should make the code
faster.

### DIFF
--- a/src/lib/device_tree/src/lib.rs
+++ b/src/lib/device_tree/src/lib.rs
@@ -109,20 +109,20 @@ impl Path {
 /// In-memory reader for Flattened Device Tree format.
 ///
 /// Does not perform any sanity checks.
-pub struct FdtReader<'a> {
-    _mem_reservation_block: SectionReader<'a>,
-    struct_block: SectionReader<'a>,
-    strings_block: SectionReader<'a>,
+pub struct FdtReader<'a, D: Driver> {
+    _mem_reservation_block: SectionReader<'a, D>,
+    struct_block: SectionReader<'a, D>,
+    strings_block: SectionReader<'a, D>,
 }
 
-fn cursor_u32(drv: &dyn Driver, cursor: &mut usize) -> Result<u32> {
+fn cursor_u32(drv: &impl Driver, cursor: &mut usize) -> Result<u32> {
     let mut data = [0; 4];
     *cursor += drv.pread(&mut data, *cursor)?;
     Ok(BigEndian::read_u32(&data))
 }
 
 // Reads a string (including null-terminator). Returns bytes read.
-fn cursor_string(drv: &dyn Driver, cursor: &mut usize, buf: &mut [u8]) -> Result<usize> {
+fn cursor_string(drv: &impl Driver, cursor: &mut usize, buf: &mut [u8]) -> Result<usize> {
     for i in 0..buf.len() {
         *cursor += drv.pread(&mut buf[i..i + 1], *cursor)?;
         if buf[i] == 0 {
@@ -137,7 +137,7 @@ fn align4(x: usize) -> usize {
 }
 
 // Reads the header of the device tree.
-fn read_fdt_header(drv: &dyn Driver) -> Result<FdtHeader> {
+fn read_fdt_header(drv: &impl Driver) -> Result<FdtHeader> {
     // 10 fields, each field 4 bytes.
     const HEADER_SIZE: usize = 10 * 4;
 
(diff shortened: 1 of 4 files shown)


src/lib/device_tree/src/lib.rs
```diff
@@ -227,7 +227,7 @@ impl<'a> Iterator for FdtIterator<'a> {
                     let value = SectionReader::new(&self.dt.struct_block, self.cursor, len);
                     self.cursor = align4(self.cursor + len);
                     // Note: This performs a copy of the whole path_buf array!
-                    return Some(Entry::Property::<'a> { path: Path { depth: self.depth + 1, len: self.len_buf, buf: self.path_buf }, value: value });
+                    return Some(Entry::Property { path: Path { depth: self.depth + 1, len: self.len_buf, buf: self.path_buf }, value: value });
```
fotonick, line 230: Clippy suggested replacing `value: value` with simply `value` in the past, but I didn't want to make a PR just for that. Could you update that here?

src/lib/device_tree/src/lib.rs
```diff
@@ -109,10 +109,10 @@ impl Path {
...
-    _mem_reservation_block: SectionReader<'a>,
-    struct_block: SectionReader<'a>,
-    strings_block: SectionReader<'a>,
+pub struct FdtReader<'a, D: Driver> {
+    _mem_reservation_block: SectionReader<'a, D>,
+    struct_block: SectionReader<'a, D>,
+    strings_block: SectionReader<'a, D>,
 }
 
 fn cursor_u32(drv: &dyn Driver, cursor: &mut usize) -> Result<u32> {
```
fotonick, line 118: Would it be in the spirit of your PR to change all dyn to impl?

fotonick (commented):
I think that this is a good idea. It doesn't seem like there will be terribly many `Driver` impls in any given build. That said, could you characterize the difference in binary size?

src/lib/device_tree/src/lib.rs
doriath: I created https://github.com/oreboot/oreboot/issues/282 to do it as separate PR (fix all existing lint errors and configure CI). Is it ok to not do this here?

src/lib/device_tree/src/lib.rs
doriath: Yes, changed all `dyn` uses inside device_tree crate to `impl`, thanks!

doriath:
The size of bootblob.bin files before and after (in bytes):

| Mainboard            | Before | After | Diff |
|----------------------|--------|-------|------|
| emulation/qemu-q35   | 65536  | 65536 | 0    |
| emulation/qemu-riscv | 13288  | 13288 | 0    |
| opentitan/crb        | 7664   | 7314  | -350 |
| sifive/hifive        | 36840  | 36338 | -502 |

I have no idea why the version with generics is smaller.

Numbers were generated with:

```
git checkout 366452794cf524f475b46a752a3b397940d880f3 && make mainboards && ~/bootblob-size.sh
git checkout b16c860f0a7863abcdb69e1e30e5097e0339555a && make mainboards && ~/bootblob-size.sh
```

Where bootblob-size.sh has following content:

```bash
#!/usr/bin/env bash

for BOOTBLOB in $(find . -name bootblob.bin)
do
  SIZE=$(stat -c"%s" $BOOTBLOB)
  if [[ $BOOTBLOB =~ \./[^/]+/[^/]+/([^/]+)/([^/]+).* ]]; then 
    echo "${BASH_REMATCH[1]}/${BASH_REMATCH[2]};${SIZE}"
  fi 
done
```

src/lib/device_tree/src/lib.rs
fotonick: Thanks for filing #282! I like your more systematic plan. I think it's OK to do the `value` replacement here or for addressing #282.

fotonick:
Thank you for running the comparison!

The results are curious, though, with some sizes changing and not others, on top of the non-zero changes being negative rather than positive. I recently noticed that `flashlayout` first reserves a fixed amount of space for each area, fills with 0xFF, then overwrites with the requested file contents. There's an explicit check that errors out if the data in the file is bigger than the space reserved. This suggests to me that the binary should always be fixed size, just padded with a variable number of 0xFF.

@rjoleary, do you have any idea how the boot blobs can change size? Maybe it's not used how I think it is?

fotonick:
Ok, figured out why I'm wrong—`bootblob.bin` is an input to `layoutflash`, not an output.

Now there's just the mystery of why no size change or shrinkage.

fotonick (approved):
But I realize that I don't even care about the negatives; it's the zeros that are confusing. Regardless, this PR is clearly a win.

PR #284, opened 2020-09-18
Makefile: separate prepare target for CI from firsttime

orangecms:
This also introduces variables to switch tooling versions.


fotonick (approved):
Looks like a nice change. Thanks!

PR #287, opened 2020-09-19
Generate the report of binary sizes in CI.

doriath:
This new script will print a simple CSV with two columns - the name of
the mainboard (two directory names) and the size of the bootblob in
bytes.

It would be nice to export this data in strucutre manner somewhere and
track over time, but for now at least we will have this information in
the azure pipeline logs.


rjoleary (approved):

fotonick:
Great idea! Thanks for doing this.

PR #289, opened 2020-09-20
Write getting started guide for new contributors.

doriath:


Documentation/getting-started.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Device Tree
+
+Read the module comment inside `src/lib/device_tree/src/lib.rs`. In oreboot,
+device tree is used for two reasons - the actual device tree of the hardware
+that we are running on (we will refer to it as `hardware device tree`) and
+device tree used to define the layout of the image that is flashed into the device
+(we will refer to it as `oreboot device tree`.
+
+## What `make run` actually does on a mainboard?
```
orangecms, line 58: Just without the question mark, since it's not formulated as a question :)

orangecms (commented):
Neat, thanks a ton! \o/

I have only skimmed through, but it looks great at first glance. :)

rminnich (commented):
if we went x86 we could completely skip the build step for riscv, and also any reference to azure, could we do that?

doriath:
About going with x86 and dropping azure:

- RISC-V is what runs on azure (CI) right now and is guaranteed to work. I found it really useful to have an example, working setup I could compare to. We would have to add x86 to CI before we update this text.
- Even if we go x86, getting azure setup was the most important part of understanding the project. README files can get out of date, but commands in CI are guaranteed to be up to date. For example, I had no idea that I should start my command with `PAYLOAD_A` to actually get `TESTTESTTEST` on the screen. Also, getting the example stdout from azure logs let me know what I should expect from my setup
- It was also important to me to know which QEMU version I should use. I first started with version from head, but it actually didn't work so I had to pin it exactly to the version from CI.

Because of the above, I would like to keep mentioning `azure-pipelines.yml` file if possible.

Documentation/getting-started.md
```diff
@@ -0,0 +1,145 @@
...
+Next, we need to install QEMU - generic machine emulator that will allow us to
+emulate running oreboot without the need of custom hardware. Below commands
+might be a little outdated, so please look at `azure-pipelines.yml` file for the
+actual commands that are used in CI.
+
+```shell
+# Go to a directory where QEMU will be fetched
+cd ~/
+# Fetch and build QEMU
+git clone https://github.com/qemu/qemu && cd qemu && git checkout <put tag from azure-pipelines.yml>
```
alistair23, line 30: I don't think you need to indicate a tag here. It would be best to just use QEMU master.

Documentation/getting-started.md
doriath: Done, thanks!

Documentation/getting-started.md
doriath: When I was starting with oreboot, HEAD master of QEMU was actually not working. For a new contributor, I wanted to increase the chance of it working for the first time (later they can experiment with other versions of QEMU).

orangecms: Is there anything wrong with just using the distro package? I'd not want to have ti build QEMU myself... and then again, why would CI need to build it? For building from source, I would definitely stick with a tagged and presumably stable release.

Documentation/getting-started.md
```diff
@@ -0,0 +1,145 @@
+# Getting Started
+
+If you would like to contribute, but don't know how to start, this page is for
+you. This guide will walk you through running Oreboot locally using QEMU
```
orangecms, line 4: oreboot should be lowercase

Documentation/getting-started.md
```diff
@@ -0,0 +1,145 @@
...
+emulation and explain the high level architecture / structure of the project.
+It also assumes the use of Debian like system.
```
orangecms, line 6: **a** Debian like system

Documentation/getting-started.md
```diff
@@ -0,0 +1,145 @@
...
+# Go to a directory where QEMU will be fetched
+cd ~/
+# Fetch and build QEMU
+git clone https://github.com/qemu/qemu && cd qemu && git checkout <put tag from azure-pipelines.yml>
+mkdir build-riscv64 && cd build-riscv64
+../configure --target-list=riscv64-softmmu
+make -j$(nproc)
+# Make the binaries available in $PATH, e.g.:
+sudo ln -s $PWD/riscv64-softmmu/qemu-system-riscv64 /usr/bin/
+sudo ln -s $PWD/qemu-img /usr/bin/
```
orangecms, line 36: I would suggest symlinking to `/usr/local/bin/` to avoid possible collisions when people install QEMU through their package manager as well.

Documentation/getting-started.md
```diff
@@ -0,0 +1,145 @@
...
+Let's go over above steps in a little more details.
+
+### Build oreboot (bootblob.bin)
+
+This is the code inside `src/mainboard/*/*/src/main.rs`. It starts with a tiny
+bit of assembler (that is included inside `main.rs`, it performs minimal
+amount of initialization and then calls into the rust program. Compared to
+other rust binaries, there is no `main` method, instead you will see `pub
+extern "C" fn ...` methods that are being called from assembler to start the
+program. The rust binary then initializes more of the hardware, prepares the
```
orangecms, line 73: **Rust** should start with an uppercase R

Documentation/getting-started.md
```diff
@@ -0,0 +1,145 @@
...
+  that it should be used to find where the payload is located that should be
+  executed, but currently the offsets are hardcoded inside the oreboot binary).
+* payload - actual binary that will be executed. It is configured with
+  environment variables, which is why the `make run` method starts with
+  `PAYLOAD_A=...`.
+
+### Run QEMU
+
+Once image is created, qemu is invoked to emulate given hardware. It flashes
+`image.bin` produced in previous step on to the device.
```
orangecms, line 115: "flashes" might be confusing here; how do you feel about "takes"?

Documentation/getting-started.md
```diff
@@ -0,0 +1,145 @@
...
+Note: There are probably more important types and crates, but those are the
+ones I got to so far in my exploration.
+
+## Important make commands
+
+Those commands are invoked from the top level `oreboot` directory:
+
+* `make format` - format all rust files in the project.
+* `make clippy` - run clippy linter on all rust files in the project.
+* `make test` - run all rust tests in the project.
```
orangecms, line 137:
```md
* `make flash` - flash to hardware, specify programmer for flashrom with `FLASHROM_PROGRAMMER`
```

Documentation/getting-started.md
alistair23: Distro packages work great, you just need to use a modern distro with v5.1.0 or newer.

Can we update this to tell users to use their distro QEMU if it's v5.1.0 or newer, otherwise build v5.1.0?

doriath: 1. This example shows RISC-V - is there prebuilt qemu for RISC-V that can be downloaded (I though this is the main reason we build it).
2. The exact version is important for RISC-V, because when I started ~month HEAD master was not working, and I needed a specific old version to have it run properly

I created #295 to setup qemu-q35 in CI. Once that is done, I think we could update this guide to just use distro package and use q35 instead of riscv. WDYT?

Documentation/getting-started.md
doriath: Good point, thanks!

Documentation/getting-started.md
doriath: Replaced all `rust` with `Rust`, thanks!

Documentation/getting-started.md
doriath: Done.

Documentation/getting-started.md
doriath: Done.

doriath (commented):
Thank you for all the comments!

Documentation/getting-started.md
alistair23: If we add q35 testing then we should test both not just 1. It doesn't make a lot of sense to remove test cases from the CI. RISC-V is stabalising quickly so oreboot should be fine to use 5.1+ for the foreseeable future.

doriath: Yes, we definitely should still test both and we should not remove RISC-V from CI. What I meant is to remove RISC-V from this guide, and just use q35 (there is no need for the user to mention both if q35 is simpler to start with). And I was suggesting to remove risc-v from this guide because of this comment https://github.com/oreboot/oreboot/pull/289#pullrequestreview-492220337 from Ron.

orangecms: In Arch, we have RISC-V QEMU packaged. 5.1.0 is the current version. See here:
https://www.archlinux.org/packages/extra/x86_64/qemu-arch-extra/

orangecms:
I'd like to propose getting the core thing in and just omit the Q35, RISC-V etc part for now, which can be moved to a follow-up PR. WDYT?

doriath:
> I'd like to propose getting the core thing in and just omit the Q35, RISC-V etc part for now, which can be moved to a follow-up PR. WDYT?

That is a good idea. Removed part of the "Setup environment" section so that we can merge it, and we can discuss risc-v/q35 in a followup.

orangecms (approved):
I like it, would just like to wait for another approval before merging. 🙂

rminnich (approved):
This is great. I would suggest getting someone who has never seen this project to try it out.

futaris:
I've seen this project, but never gotten around to setting up a build process, or use oreboot.

I use x86 coreboot in my dayjob and fiddle with some RISC-V FPGA stuff outside of work.

futaris:
Ran through everything. It all worked on a Ubuntu 18.04.5 LTS x86-64 system.

Might be worth mentioning system / disk space requirements. 2GiB space is probably more than enough.

The only other thing that I would test, would be what other build packages might be needed before the git pull.  Either create a VM with a fresh install of a Debian system, or a docker or something similar. Does the user just need to `apt install git` or install anything else?

I haven't tried anything with Qemu, as those instructions aren't there (yet)? I did build oreboot from 
`/src/mainboard/emulation/qemu-q35` without any issues though.

Feel free to ping me on the Q35 / RISC-V running oreboot in Qemu follow-on patch, for review. Either on github or https://slack.osfw.dev/ #oreboot ...

PR #294, opened 2020-09-23
Add and convert Coffeelake FSP headers to Rust

rjoleary:
The FSP header files depend on edk2 header files, so both are added as
submodules.

To checkout submodules while cloning:

  git clone --recurse-submodules https://github.com/oreboot/oreboot

To update submodules:

  git submodule update --init

The conversion takes place during a build step.

Only structs and constants are included during the conversion. Thus,
this will not impact the safety of Rust and save us from the error-prone
task of converting by hand.

Types are explicitly included via the build.rs file.

For docs, run `cargo doc --open`.


doriath:
I think you might need to add your crate to `BROKEN_CRATES_TO_TEST` variable inside `Makefile` to fix the broken CI build.

There are many crates in this project for which `cargo test` fails, and I will be analyzing them soon, but for now if new crate is added and doesn't work just add it to broken list.

rjoleary:
> I think you might need to add your crate to `BROKEN_CRATES_TO_TEST` variable inside `Makefile` to fix the broken CI build.
> 
> There are many crates in this project for which `cargo test` fails, and I will be analyzing them soon, but for now if new crate is added and doesn't work just add it to broken list.

Thanks! I think it's my fault this is broken. I need to add another dependency to `make firstime` from https://rust-lang.github.io/rust-bindgen/requirements.html#debian-based-linuxes

src/vendorcode/fsp/coffeelake/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "coffeelake"
```
fotonick, line 2: The typical convention is that raw bindgen wrapper crates have `-sys` appended to the crate names. It leaves room for a more idiomatic Rust interface to be exposed without the `-sys`. May I also suggest `fsp` in the crate name somewhere in case we ever want to publish this separately?

rjoleary: Thanks!

fotonick (approved):

orangecms (approved):

PR #296, opened 2020-09-23
Add more tests for device tree reader.

### LINKED ISSUE #155: fdt parser fails if value is length 0

### LINKED ISSUE #156: Write unit tests for fdt parser
Julian has already started working on this. I believe this will be the first unit test.

doriath:
While writing tests, I found that there is a bug in SliceReader (it
always returns the size of the input buffer instead of the number of
bytes actually read).

Closes #156 - existing tests cover various cases, including nested
nodes, different type of properties (empty, number, string) and few
error conditions.

Closes #155 - added test cases for empty properties and shows it works
(though because of Driver API, when property is empty the Driver returns
EOF error, but that is WAI).

### DIFF
--- a/src/drivers/wrappers/tests/tests.rs
+++ b/src/drivers/wrappers/tests/tests.rs
@@ -32,21 +32,30 @@ fn test_slice_reader_returns_zero_for_empty_buf() {
 }
 
 #[test]
-#[should_panic]
+fn test_slice_reader_returns_number_of_bytes_read() {
+    let data = [1, 2, 3, 4, 5];
+    let driver = SliceReader::new(&data);
+    let mut buf = [0; 10];
+
+    assert_eq!(5, driver.pread(&mut buf, 0).unwrap());
+    assert_eq!(&buf[..5], &data);
+    assert_eq!(&buf[5..], &[0; 5]);
+}
+
+#[test]
 fn test_slice_reader_returns_eof() {
     let data = [1, 2, 3, 4, 5];
     let driver = SliceReader::new(&data);
     let mut buf = [0; 5];
 
-    driver.pread(&mut buf, 5).unwrap();
+    assert_eq!(driver.pread(&mut buf, 5).err(), Some("EOF"));
 }
 
 #[test]
-#[should_panic]
 fn test_slice_reader_returns_eof_with_empty_buf() {
     let data = [1, 2, 3, 4, 5];
     let driver = SliceReader::new(&data);
     let mut buf = [0; 0];
 
-    driver.pread(&mut buf, 5).unwrap();
+    assert_eq!(driver.pread(&mut buf, 5).err(), Some("EOF"));
 }
--- a/src/lib/device_tree/tests/integration_test.rs
+++ b/src/lib/device_tree/tests/integration_test.rs
@@ -37,7 +37,12 @@ fn dts_to_dtb(dts: &str) -> std::vec::Vec<u8> {
 
 fn read_all(d: &impl Driver) -> std::vec::Vec<u8> {
     let mut data = [0; 500];
-    let size = d.pread(&mut data, 0).unwrap();
+    let size = match d.pread(&mut data, 0) {
+        Ok(x) => Ok(x),
+        Err("EOF") => Ok(0),
+        Err(other) => Err(other),
+    }
+    .unwrap();
     let mut result = std::vec::Vec::new();
     result.extend_from_slice(&data[0..size]);
     return result;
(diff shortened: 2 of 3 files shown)


src/lib/device_tree/tests/integration_test.rs
```diff
@@ -71,3 +76,78 @@ fn test_reads_properties() {
...
+};"#,
+    );
+    // Magic is a first number
+    data[0] = 1;
+    let slice_reader = &SliceReader::new(&data);
+    FdtReader::new(slice_reader).unwrap();
+}
+
+#[test]
+#[should_panic(expected = "not enough data to read device tree header")]
```
fotonick, line 142: Probably out of scope for this PR, but I find it very un-Rusty to panic instead of returning a `Result::Err`. I would imagine that there are contexts in which this could be recoverable.

src/drivers/wrappers/tests/tests.rs
```diff
@@ -31,6 +31,15 @@ fn test_slice_reader_returns_zero_for_empty_buf() {
...
 }
 
+#[test]
+fn test_slice_reader_returns_number_of_bytes_read() {
+    let data = [1, 2, 3, 4, 5];
+    let driver = SliceReader::new(&data);
+    let mut buf = [0; 10];
+
+    assert_eq!(5, driver.pread(&mut buf, 0).unwrap());
+}
```
fotonick, line 41: Worth also testing contents?

    assert_eq!(&buf[..5], &data);
    assert_eq!(&buf[5..], &[0; 5]);

src/lib/device_tree/tests/integration_test.rs
```diff
@@ -37,7 +37,12 @@ fn dts_to_dtb(dts: &str) -> std::vec::Vec<u8> {
 
 fn read_all(d: &impl Driver) -> std::vec::Vec<u8> {
     let mut data = [0; 500];
-    let size = d.pread(&mut data, 0).unwrap();
+    let size = match d.pread(&mut data, 0) {
+        Ok(x) => Ok(x),
+        Err("EOF") => Ok(0),
+        y => y,
```
fotonick, line 43: I had to think about this a while before understanding that the `y => y` match arm was to catch any `Err` strings other than `"EOF"`. Normally I think of Ok + Err cases being the complete set, so this stood out to me. How would you feel about making this `Err(other) => Err(other)` for human clarity?

fotonick (dismissed):
Suggestions to fix at your discretion. Thanks for adding these tests.

src/drivers/wrappers/tests/tests.rs
doriath: That is a good idea, thanks!

src/lib/device_tree/tests/integration_test.rs
doriath: I agree it is much more readable, thanks!

src/lib/device_tree/tests/integration_test.rs
doriath: I converted `should_panic` tests into

```rust
assert_eq!(func().err, Some("...."))
```

To verify that function returns an expected error on invalid input. I think it is also cleaner now that we expect error in specific place, instead of catching any panic in this function.

src/lib/device_tree/tests/integration_test.rs
```diff
@@ -71,3 +76,76 @@ fn test_reads_properties() {
...
+        r#"
+/dts-v1/;
+/ {
+    #address-cells = "ok";
+};"#,
+    );
+    // Magic is a first number
+    data[0] = 1;
+    let slice_reader = &SliceReader::new(&data);
+    assert_eq!(FdtReader::new(slice_reader).err(), Some("invalid magic in device tree header"));
```
fotonick, line 137: Nit: More direct and idiomatic would be `assert_eq!(FdtReader::new(slice_reader), Err("invalid magic in device tree header"));`. It is a confusing digression to convert from `Result` to `Option` with `.err()`.

src/lib/device_tree/tests/integration_test.rs
```diff
@@ -71,3 +76,76 @@ fn test_reads_properties() {
...
+fn test_returns_error_when_header_is_too_short() {
+    let data = dts_to_dtb(
+        r#"
+/dts-v1/;
+/ {
+    #address-cells = "ok";
+};"#,
+    );
+    let slice_reader = &SliceReader::new(&data[0..4]);
+    assert_eq!(FdtReader::new(slice_reader).err(), Some("not enough data to read device tree header"));
```
fotonick, line 150: Same as above.

src/lib/device_tree/tests/integration_test.rs
doriath: I believe that would require FdtReader to derive .Debug, PartialEq traits so that assert_eq can print the value in case it is not an error, as well as compare two FdtReaders. I saw using "err()" and comparing with Option as one workaround for it.

src/lib/device_tree/tests/integration_test.rs
doriath: As above.

src/lib/device_tree/tests/integration_test.rs
fotonick: Oh, I see. Yes. The full type is `Result<FdtReader, &str>` and calling `.err()` allows you to discard the first type so you can compare based on the `&str` alone. Thanks for clarifying.

fotonick (approved):

PR #298, opened 2020-09-24
Use github.com/oreboot/simbin for qemu

gevorgyana:

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -45,12 +45,9 @@ steps:
   displayName: 'Build QEMU q35 board for x86'
 - script: |
     set -e
-    git clone https://github.com/qemu/qemu && cd qemu && git checkout v5.1.0
-    mkdir build-riscv64 && cd build-riscv64
-    ../configure --target-list=riscv64-softmmu
-    make -j16
-    sudo ln -s $PWD/riscv64-softmmu/qemu-system-riscv64 /usr/bin/
-    sudo ln -s $PWD/qemu-img /usr/bin/
+    git clone https://github.com/gevorgyana/simbin.git && cd simbin
+    ./build.sh
+    sudo cp qemu/build-riscv64/riscv64-softmmu/qemu-system-riscv64 /usr/bin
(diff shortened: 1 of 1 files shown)


gevorgyana:
If @rminnich merges this https://github.com/oreboot/simbin/pull/1, then we can checkout that repository directly in azure-yaml.

gevorgyana:
Having a QEMU binary would eliminate the need for compiling it, which leads to less time spent in CI, at the (small) cost of updating the binary from the upstream from time to time.

azure-pipelines.yml
```diff
@@ -51,6 +51,7 @@ steps:
     make -j16
     sudo ln -s $PWD/riscv64-softmmu/qemu-system-riscv64 /usr/bin/
     sudo ln -s $PWD/qemu-img /usr/bin/
+
```
rjoleary, line 54: All I see inserting a blank line

azure-pipelines.yml
```diff
@@ -45,12 +45,9 @@ steps:
...
     set -e
-    git clone https://github.com/qemu/qemu && cd qemu && git checkout v5.1.0
-    mkdir build-riscv64 && cd build-riscv64
-    ../configure --target-list=riscv64-softmmu
-    make -j16
-    sudo ln -s $PWD/riscv64-softmmu/qemu-system-riscv64 /usr/bin/
-    sudo ln -s $PWD/qemu-img /usr/bin/
+    git clone https://github.com/gevorgyana/simbin.git && cd simbin
+    ./build.sh
+    sudo cp qemu/build-riscv64/riscv64-softmmu/qemu-system-riscv64 /usr/bin
```
gevorgyana, line 50: For now, this builds qemu from source, done here to prove that it works. When this oreboot/simbin#1 is merged, we will just `git checkout https://github.com/oreboot/simbin.git` and `sudo cp` the binary to `usr/bin`. Anyway we should probably wait for this oreboot/simbin#1.

rjoleary:
I'll have an updated version in another PR...

Issue #295, opened 2020-09-23
Setup qemu-q35 on CI


doriath:
The goal is to improve #289 getting started guide to use x86 instead of RISC-V.

- Run qemu-q35 on CI similar how RISC-V is running now. The goal is to ensure it always works and if someone starts with the project, they can look at the working example.
- Ensure qemu-q35 actually works (currently if you run it, it panics, but maybe it is because there is no payload and I am not sure which payload will work for it)
- Once #289 is in, update it to use x86 instead of RISC-V to setup the environment

Issue #100, opened 2019-08-11
Use caching to speed up CI


rjoleary:

fotonick:
Consider a shallow git clone (`git clone -–depth 1`) to speed that phase up.

Azure is pouring a ton of time into building QEMU 4.2. And are you willing to point Debian to the testing package repo? I see that [4.2 is available on testing](https://packages.debian.org/search?keywords=qemu).

mpictor:
If people are OK with using a pre-built binary, one solution I've used (that doesn't involve storage in git or use of a paid blobstore) is to upload a tarball to github as an asset for some particular release, such as the very first. Then, just download the tarball on demand.

I've done that with a custom compiler tarball that bazel [downloaded](https://github.com/mpictor/bazel-amd64-crosstool/blob/e6c5e492b3ac2f8f521ac316f731782e38eb9cbd/tools/deps.bzl#L9). Note that I always downloaded from the same release; to download from the latest, you'd have to upload again with each new release.

fotonick:
I spent some time looking at related ideas in PR #255. Our Azure CI provides artifact save and restore features. I found that caching the firmware binaries saves surprisingly little time because the Rust toolchain takes so much time to set up. (Rust toolchain is needed for both the build and test phases as we've set it up today; could be removed from test phase with some work, I think.) Caching the Rust toolchain itself saves surprisingly little time because of way-slower-than-I-expected network speeds to the cache (~10 MBps, IIRC).

I somehow didn't look into caching the QEMU build. I really should have, since it's ~4 minutes of the CI build, IIRC, and probably not such a huge binary size. That's a good idea.

mpictor:
If qemu is being built for a single well-defined target, you can also speed the build by disabling unneeded things. IIRC [this](https://github.com/purecloudlabs/gprovision/blob/master/build/qemu/Dockerfile#L87) made quite a difference.

mpictor:
If I change the qemu build step to this
```bash
    set -e
    git clone -b v5.1.0 https://github.com/qemu/qemu --depth 1 && pushd qemu
    mkdir build && cd build
[...]

PR #299, opened 2020-09-25
azure-pipelines: build qemu with x86_64 support, test qemu-q35 board

mpictor:
There's at least one other PR related to qemu (#298), but that one uses a prebuilt binary for risc-v where this one adds x86_64 support.

Thought I'd create the PR anyway and see how well azure liked it. Related: #100, #295.

Signed-off-by: Mark Pictor <mpictor@gmail.com>


fotonick:
Thanks for this!

You probably saw that CI failed with `ERROR: VirtFS requires libcap-ng devel and libattr devel` ([log](https://dev.azure.com/azure0427/Oreboot%20Pipeline/_build/results?buildId=1133&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=4a279652-4b54-5341-2558-7da0cc58b911&l=32)). Perhaps you have some local packages installed that are not listed in the [`debiansysprepare` Make target](https://github.com/oreboot/oreboot/blob/ffed7b8498593d71fd5f83e4ad3e0cddbf023b8d/Makefile#L44).

It's really useful to create PRs to get CI to run on your commits, but I might recommend that if it isn't ready to review yet, then mark a PR as draft.

mpictor:
> Perhaps you have some local packages installed that are not listed in the debiansysprepare Make target.

I use gentoo locally, so it can be a bit of a guessing game what will be needed; I could use a container to cut down on the discovery.

> It's really useful to create PRs to get CI to run on your commits, but I might recommend that if it isn't ready to review yet, then mark a PR as draft.

Will do; for some reason I assumed that'd prevent CI from running.

mpictor:
@fotonick did you have any luck with caching? Pretty sure I followed their instructions but the step

```yml
- task: Cache@2
  inputs:
    # Note that caches are immutable. To change, use a different name (e.g. qemu-bins2-...) or
    # update QVER. Cache expires after 7 days of no use.
    key: 'qemu-bin-v${{ variables.QVER }}'
    path: $(Pipeline.Workspace)/qemu-${{ variables.QVER }}-bin.tar.xz
    cacheHitVar: QEMU_CACHE_RESTORED
  displayName: Get QEMU cache
```
fails:
```
Starting: Get QEMU cache
==============================================================================
Task         : Cache
Description  : Cache files between runs
Version      : 2.0.1
Author       : Microsoft Corporation
Help         : https://aka.ms/pipeline-caching-docs
==============================================================================
Resolving key:
 - qemu-bin-v5.1.0 [file] (not found)
##[error]One or more errors occurred. (File not found: qemu-bin-v5.1.0)
Finishing: Get QEMU cache
```
[CI build link](https://dev.azure.com/azure0427/Oreboot%20Pipeline/_build/results?buildId=1152&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=3fa863a0-64d8-5210-739a-0743ce195f76&l=11)

\<rant\>
Their description of keys seems magical, and not in a good way. Segments of the key can be plain text, or might be replaced by the hash of a file by the same name, and apparently can match a prefix instead of requiring an exact match. For that gem, scroll to just above `Cache Isolation and Security` [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops).
\</rant\>

fotonick:
@mpictor Thanks for digging deeper. I think what I got working was publish/download rather than caching, mostly because I found caching confusing. =) From their description of the different purposes, this is definitely a case where the caching description fits better—it doesn't need to be a durable artifact and we're not concerned with having an authoritative version.

Your guess at using caching seems reasonable to me. I have no particular ideas on what might be wrong. Sorry I'm not more useful.

mpictor:
A comment by @rminnich on slack made me think about containers - and that building once in a container made more sense. So I've been fiddling with that. I got it to build, but it's rather big. Your mention of slow transfer speeds makes me wonder if a container transfer would be affected the same way as the publish/download or whatever you tried.

Using rust's official container might cut down on size, but I think that'd mean translation of `rustup` commands into something else - something that I have even less experience with :)

My work on a dockerfile can be seen on [this branch](https://github.com/mpictor/oreboot/commits/docker-ci) (currently only one commit). I've done some cleanup since I last tried to build it; it's possible there are typos, but it shouldn't be too far off.

fotonick:
I played with Docker containers [a bit](https://github.com/oreboot/oreboot/pull/255#issuecomment-622804898). Unfortunately the Rust toolchain is just enormous (~1 GB), so I dropped this line of inquiry. I saw that the official Rust toolchain is comparable size to what I got building the image myself. That said, you may succeed where I did not. The ultimate arbiter will be total CI wall clock time, tempered by our judgment about additional complexity.

mpictor:
Closing, as I don't see myself having time to get this working. I'll leave the [branch](https://github.com/mpictor/oreboot/tree/azure-x86_64) up in case someone else wishes to take a look.

orangecms:
A little update here: We got a Docker setup to build an oreboot image with prepared scripts and patches from @daym, based on the upstream Rust nightly image. However, `dtc` is quite tightly integrated into our build toolchain. There are two ways to go about it:
- install the `device-tree-compiler` package into the image - that makes it not fully reproducible and a bit larger; it is not too bad though since the Rust toolchain is mostly self-contained in the image afaict
- redesign the build system such that `dtc` can be invoked from outside; though then you'd still need it from somewhere again

As of now, the oreboot image for x86 / q35 is being built in CI: https://github.com/oreboot/oreboot/blob/master/azure-pipelines.yml#L41
The current implementation only worked with a bzImage when I tried yesterday. Though it did very well. I used https://github.com/oreboot/binaries/blob/master/rminnich/romecrb/orebootkernel. We can use this very kernel image for a check in CI. It contains `u-root` and a few utilities.

orangecms:
Here is a usable `Dockerfile`:

```Dockerfile
FROM rustlang/rust:nightly AS rust-nightly-with-rust-src
RUN rustup override set nightly-2020-10-25
RUN rustup component add rust-src
RUN apt-get update
RUN apt-get install device-tree-compiler
```

Build it with a tag for local use: `docker build -t rust-nightly-extra .`

To build an oreboot image, from the oreboot root directory, use this script, say `build.sh`:
```bash
#!/bin/sh

docker run --rm \
  -e RUST_TARGET_PATH="${RUST_TARGET_PATH}" \
  -e TARGET_DIR="${TARGET_DIR}" \
  -e PAYLOAD_A="${PAYLOAD_A}" \
  -e USER="${USER}" \
  --user "$(id -u)":"$(id -g)" \
  -v "${PWD}:${PWD}" -w "${p}" \
  rust-nightly-extra:latest "$@"
```

Now we need to set the env var `p` to our working directory, and pass the `make` variable `OBJCOPY` to `objcopy` to build:

```sh
p="$(pwd)" PAYLOAD_A="$(pwd)"/orebootkernel ./build.sh make -C src/mainboard/emulation/qemu-q35/ OBJCOPY=objcopy
```

And we get the oreboot image, which we can run with QEMU; just mind giving it enough memory:

```sh
qemu-system-x86_64 -m 2g -machine q35 -bios src/mainboard/emulation/qemu-q35/target/x86_64-unknown-none/release/image.bin -serial stdio
```

daym:
> There are two ways to go about it:

There are three ways to go about it.

> * install the `device-tree-compiler` package into out derived Docker image - that makes it not fully reproducible and a bit larger; it is not too bad though since the Rust toolchain is mostly self-contained in the image afaict

Yeah, this one I would not recommend.

> * redesign the build system such that `dtc` can be invoked from outside; though then you'd still need it from somewhere again

It can do that now. GNU `make` only invokes a recipe if the output file is older than the input file.

So that means you can just, as the very first step, put the output (dtb) file where it's supposed to be and `make` won't touch it anymore.

See https://github.com/daym/docker-oreboot/blob/main/build - it works fine.

The third way would be to just add the dtb file to the oreboot git repo.

The substitution of the `$(...)` variables in the dts is done not by dtc (!!!) but by oreboot later in the loaded dtb, so there is no problem just making it static (the dtb contents will "never" change anyway--even now).

orangecms:
Yea, option 4 sounds good, let's go for it! :-)

daym:
Using https://git.kernel.org/pub/scm/utils/dtc/dtc.git/snapshot/dtc-1.6.0.tar.gz and in-rust-nightly, it says flex and bison is missing in the container :(

So that doesn't work either.

Since the docker container as I use it self-destructs after each invocation of in-rust-nightly, I guess it's possible to do the apt-get install device-tree-compiler and then invoke it once and then it will be removed anyway.

What I wouldn't do is put that in the Dockerfile.

rjoleary:
@orangecms Is going to pickup the caching.

PR #302, opened 2020-10-02
Remove non-necessary Makefile.toml

gevorgyana:
Most probably, the author
of 12c6b52cb42e10c forgot to do it.

As stated in that commit, the goal was to remove usages of cargo-make.

Signed-off-by: Artyom Gevorgyan <artemhevorhian@gmail.com>


orangecms (dismissed):
looks like a leftover, yup :)

fotonick:
If it's convenient, would you remove `src/mainboard/opentitan/crb/Makefile.toml`, as well? Looks like the same deal.

doriath:
nit: what do you think about folding both commits into one?

gevorgyana:
@doriath I thought we do squashed merges into master, okay.

orangecms (approved):

doriath:
I believe we do `rebase` by default (in `.github/auto-merge.yml` there is `mergeMethod: rebase`.

orangecms:
Correct. When there is a pile of patches and auto-merge kicks in, they would end up just like that in master. Also, if someone were to merge a set of patches, then they would need to write a new message when squashing, so yea... Just prepare the patches so that they can be taken as-is, that makes it easy to handl them. :)

gevorgyana:
@doriath @orangecms Thanks for information :+1:

PR #304, opened 2020-10-04
Refactor FdtIterator to use streaming iterator API.

### LINKED ISSUE #74: Use streaming iterator to iterate over device tree

doriath:
Closes #74.

We are dropping the core::Iterator trait from `FdtIterator` as we never
actually benefited from implementing it. Implementing next with custom
lifetimes allows us to make it more performant (`Entry` can reference
internal buffers in `FdtIterator`).

I am also adding error handling to the `FdtIterator` following
`std::io::Lines` approach. I also removed the use of unsafe from device
tree crate (we check if string is indeed utf8 and return error if it is
not).

I also slightly changed the API of Entry to include both StartNode and
EndNode. This is following similar pattern to StAX XML parsers and
allows users of the iterator to write more composable code.

### DIFF
--- a/src/lib/device_tree/src/lib.rs
+++ b/src/lib/device_tree/src/lib.rs
@@ -23,6 +23,7 @@ extern crate num_derive;
 
 use byteorder::{BigEndian, ByteOrder};
 use core::fmt;
+use core::fmt::Write;
 use model::{Driver, Result};
 use wrappers::SectionReader;
 
@@ -234,6 +235,37 @@ impl<'a, D: Driver> FdtIterator<'a, D> {
         }
         Err("EOF")
     }
+}
+
+/// Reads the device tree in FDT format from given driver and writes it in human readable form to
+/// given writer.
+pub fn print_fdt(fdt: &mut impl Driver, w: &mut print::WriteTo) -> Result<()> {
+    let reader = FdtReader::new(fdt)?;
+    let mut iter = reader.walk();
+    let mut depth = 0;
+    while let Some(entry) = iter.next() {
+        let entry = entry?;
+        match entry {
+            Entry::StartNode { name } => {
+                depth += 1;
+                write!(w, "{:depth$}{}\r\n", "", name, depth = depth * 2).unwrap();
+            }
+            Entry::EndNode {} => {
+                depth -= 1;
+            }
+            Entry::Property { name, value: v } => {
+                let buf = &mut [0; 1024];
+                let len = match v.pread(buf, 0) {
+                    Ok(x) => x,
+                    Err("EOF") => 0,
+                    Err(y) => return Err(y),
+                };
+                let val = infer_type(&buf[..len]);
+                write!(w, "{:depth$}{} = {}\r\n", "", name, val, depth = depth * 2,).unwrap();
+            }
+        }
+    }
+    Ok(())
 }
 
 pub enum Entry<'a, D: Driver> {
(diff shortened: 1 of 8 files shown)


src/lib/device_tree/src/lib.rs
```diff
@@ -16,7 +16,6 @@
 //! For more information about Device Tree Standard, see https://www.devicetree.org/.
 
 #![no_std]
-#![deny(warnings)]
```
rjoleary, line 19: Is it possible to leave this in?

src/lib/device_tree/src/lib.rs
```diff
@@ -171,55 +170,75 @@ impl<'a, D: Driver> FdtReader<'a, D> {
...
                 Some(Token::Prop) => {
                     let len = cursor_u32(&self.dt.struct_block, &mut self.cursor).unwrap() as usize;
-                    let mut nameoff = cursor_u32(&self.dt.struct_block, &mut self.cursor).unwrap() as usize;
-                    let name_len = cursor_string(&self.dt.strings_block, &mut nameoff, &mut self.name_buf[..]).unwrap() as usize;
+                    let mut name_off = cursor_u32(&self.dt.struct_block, &mut self.cursor).unwrap() as usize;
+                    let name_len = cursor_string(&self.dt.strings_block, &mut name_off, &mut self.name_buf[..]).unwrap() as usize;
                     let value = SectionReader::new(&self.dt.struct_block, self.cursor, len);
                     self.cursor = align4(self.cursor + len);
-                    return Some(Entry::Property { path: Path { depth: self.depth + 1, len: name_len, buf: self.name_buf }, value });
+                    match core::str::from_utf8(&self.name_buf[..name_len]) {
```
rjoleary, line 204: Do you know what the spec says about string encoding? I skimmed through and couldn't see anything.

src/mainboard/sifive/hifive/src/main.rs
```diff
@@ -176,20 +176,28 @@ fn test_ddr(addr: *mut u32, size: usize, w: &mut print::WriteTo) -> Result<(), (
 
 // TODO: move out of mainboard
```
rjoleary, line 177: Would this be a good time to move this to the device_tree crate?

src/lib/device_tree/src/lib.rs
```diff
@@ -171,55 +170,75 @@ impl<'a, D: Driver> FdtReader<'a, D> {
...
+                        Ok(name) => return Some(Ok(Entry::Property { name, value })),
+                        Err(_) => return Some(Err("property name is not valid uft8")),
+                    }
                 }
                 Some(Token::Nop) => continue,
                 Some(Token::End) => return None,
                 None => panic!("unexpected token"),
```
rjoleary, line 211: Now that the function returns an error, can you return an error here instead of panic?

src/lib/device_tree/src/lib.rs
doriath: Yes, reverted the change (I forgot to revert it after development)

src/lib/device_tree/src/lib.rs
doriath: Done (note: there are more `unwrap` calls in this crate that i am planning to go over and fix in separate PR).

src/mainboard/sifive/hifive/src/main.rs
doriath: Done (I decided to do this as separate commit in this PR).

rjoleary (approved):

PR #306, opened 2020-10-13
Add unit test to i8250 driver and remove unused code

Rwantare:
Unfortunately i8250 is port mapped so I am unable to use register blocks.
I still managed to add some unit test to check that the driver initialises the UART.

### DIFF
--- a/src/drivers/uart/src/i8250.rs
+++ b/src/drivers/uart/src/i8250.rs
@@ -74,7 +74,7 @@ impl<'a> Driver for I8250<'a> {
     fn pread(&self, data: &mut [u8], _offset: usize) -> Result<usize> {
         for c in data.iter_mut() {
             let mut s = [0u8; 1];
-            while self.poll_status(1, 1) {}
+            while !self.poll_status(1, 1) {}
             self.d.pread(&mut s, self.base).unwrap();
             *c = s[0];
         }
@@ -84,7 +84,7 @@ impl<'a> Driver for I8250<'a> {
     fn pwrite(&mut self, data: &[u8], _offset: usize) -> Result<usize> {
         for (_i, &c) in data.iter().enumerate() {
             // Poll the status for long enough to let a char out; then push it out anyway.
-            self.poll_status(2, 2);
+            while !self.poll_status(0x20, 0x20) && !self.poll_status(0x40, 0x40) {}
             let mut s = [0u8; 1];
             s[0] = c;
             self.d.pwrite(&s, self.base).unwrap();
--- a/src/drivers/uart/src/i8250.rs
+++ b/src/drivers/uart/src/i8250.rs
@@ -16,7 +16,7 @@ pub struct RegisterBlock {
 */
 pub struct I8250<'a> {
     base: usize,
-    //baud: u32,
+    baud: u32,
     d: &'a mut dyn Driver,
 }
 
@@ -31,8 +31,9 @@ pub struct I8250<'a> {
 // }
 
 impl<'a> I8250<'a> {
-    pub fn new(base: u16, _baudrate: u32, d: &'a mut dyn Driver) -> I8250<'a> {
-        I8250 { base: base as usize, d: d }
+    // why is base a usize? for mmio 8250.
+    pub fn new(base: usize, baud: u32, d: &'a mut dyn Driver) -> I8250<'a> {
+        I8250 { base: base, baud: baud, d: d }
     }
 
     /// Returns a pointer to the register block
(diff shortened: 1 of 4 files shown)


src/drivers/uart/src/i8250.rs
```diff
@@ -95,43 +99,74 @@ impl<'a> Driver for I8250<'a> {
...
-//     ]
-// }
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /*
+     * UART pushing into a vec
+     */
+    extern crate heapless;
```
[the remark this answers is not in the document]
Rwantare: I need the heapless Vec because of no_std

src/drivers/uart/src/i8250.rs
```diff
@@ -95,43 +99,74 @@ impl<'a> Driver for I8250<'a> {
...
+            MockPort { ldata: v }
+        }
+    }
+
+    impl<'a> Driver for MockPort<'a> {
+        fn init(&mut self) -> Result<()> {
+            Ok(())
+        }
+
+        fn pread(&self, _data: &mut [u8], _offset: usize) -> Result<usize> {
```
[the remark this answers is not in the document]
Rwantare: Yes, UART init isn't reading anything, but I can implement it for later.

src/drivers/uart/src/i8250.rs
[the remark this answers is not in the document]
GanShun: damn that's not great if our tests are subject to the same constraints as our code because they're in the same file and package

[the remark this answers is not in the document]
Rwantare: It makes the test more faithful though

rminnich (dismissed):

src/drivers/uart/src/lib.rs
```diff
@@ -15,3 +15,8 @@ pub mod opentitan;
 pub mod pl011;
 #[cfg(feature = "sifive")]
 pub mod sifive;
+
+/* Calculate divisor. Do not floor but round to nearest integer. */
```
rjoleary, line 19: Use the triple-style comment here for function-level documentation `///`.

src/drivers/uart/src/i8250.rs
```diff
@@ -55,26 +27,58 @@ impl<'a> I8250<'a> {
     }
 }
 
+#[allow(dead_code)]
```
rjoleary, line 30: Why allow dead code?

src/drivers/uart/src/lib.rs
```diff
@@ -15,3 +15,8 @@ pub mod opentitan;
...
+pub fn uart_baudrate_divisor(baudrate: usize, refclk: usize, oversample: usize) -> usize {
```
rjoleary, line 20: This function looks like it never gets used.

src/drivers/uart/src/i8250.rs
```diff
@@ -1,47 +1,19 @@
...
-
-//     fn deref(&self) -> &Self::Target {
-//         unsafe { &*self.ptr() }
-//     }
-// }
-
 impl<'a> I8250<'a> {
-    pub fn new(base: u16, _baudrate: u32, d: &'a mut dyn Driver) -> I8250<'a> {
-        I8250 { base: base as usize, d: d }
+    // why is base a usize? for mmio 8250.
```
rjoleary, line 10: What is the question? https://doc.rust-lang.org/std/primitive.usize.html "The size of this primitive is how many bytes it takes to reference any location in memory. For example, on a 32 bit target, this is 4 bytes and on a 64 bit target, this is 8 bytes."

src/drivers/uart/src/i8250.rs
```diff
@@ -95,43 +99,78 @@ impl<'a> Driver for I8250<'a> {
...
+            Ok(data.len())
+        }
+
+        fn shutdown(&mut self) {}
+    }
+
+    const FCR: usize = 0x02; // FIFO Control Register                0b0010 W
+    const LCR: usize = 0x03; // Line Control Register                0b0011 RW
+                             // const DLL: usize = 0x00; // Divisor Latch Low Byte               0      RW
+                             // const DLH: usize = 0x01; // Divisor Latch High Byte              0x0001 RW
```
rjoleary, line 152: Remove this commented code if it's not used.

src/drivers/uart/src/lib.rs
Rwantare: It doesn't. It is also seems unnecessary because the maximum baudrate is used when possible.

src/drivers/uart/src/i8250.rs
Rwantare: Done

src/drivers/uart/src/i8250.rs
Rwantare: Allow dead code means the UART constants can be included even though a lot of them are unused. Not sure whether I should remove them.

GanShun (approved):

PR #307, opened 2020-10-20
add bios table support from xhype

rminnich:


src/arch/x86/x86_64/src/bios.rs
```diff
@@ -0,0 +1,380 @@
...
+ * Copyright (c) 2000 - 2014 Intel Corporation
+ *
+ * Compilation of "vmm_acpi_dsdt.dsl" - Fri Apr  1 13:34:26 2016
+ *
+ */
+
+/*
+ *
+ * Based on the example at osdev wiki wiki.osdev.org/AML,
+ * and the other example in http://www.acpi.info/DOWNLOADS/ACPI_5_Errata%20A.pdf
```
orangecms, line 227: Could you please add https://uefi.org/acpi/specs to save people some time digging? Things got moved around and the former ACPI website now just redirects to Intel. 🤡

src/arch/x86/x86_64/src/bios.rs
```diff
@@ -0,0 +1,380 @@
...
+    }
+    tot as u8
+}
+
+// I still don't know why these were different, I thought it was the
+// same algorithm.
+#[inline]
+fn acpi_tb_checksum(start: usize, end: usize) -> u8 {
+    // I don't understand this one
+    // data.iter().map(|x| Wrapping(*x)).sum::<Wrapping<u8>>().0
```
orangecms, line 267: That is functional programming style, and though and don't know Rust that well yet nor the data structures here, this is what I read - in very short:
- get a different view on `data` by calling `iter()` on it; looks like that converts the underlying data structure to something that can be operated on
- apply a `map` - for every element in the previous set, create a new one; i.e., the number of elements stays the same. `Wrapping(*x)` seems to covert to another data structure again
- call the `sum()` function on the set, dead simple; a built-in function that knows how to operate on a set of numbers

The final notation thing looks like typing shenanigans again, and whatever the `.0` does - no idea.

I hope I could provide some ideas. :)

And now for the outcome: If the two implementations give different sums, well... 🙃

orangecms: In other words: If Rust was super duper smart, you could just call `data.sum()`.

fotonick: The whole `Wrapping` bit is to specify that addition should wrap on overflow instead of saturating or (default) panic on overflow.

The `.0` is to extract from `Wrapping(u8)` to `u8`.

An alternative to using a `Wrapping` type is to call `wrapping_add()`. It might be more readable here than this extra wrapping and unwrapping (in the type sense) that we see here.

rminnich: Wrapping is an add that won't overflow. 
the .0 is a notation meaning 'first element in tuple' I believe.

Doesn't matter much, it's more complex than can be done in nostd.

Rwantare (approved):

src/arch/x86/x86_64/src/bios.rs
```diff
@@ -0,0 +1,380 @@
...
+    pub oem_table_id: [u8; 8],      /* ASCII OEM table identification */
+    pub oem_revision: u32,          /* OEM revision number */
+    pub asl_compiler_id: [u8; 4],   /* ASCII ASL compiler vendor ID */
+    pub asl_compiler_revision: u32, /* ASL compiler version */
+}
+pub const ACPI_TABLE_HEADER_CHECKSUM_OFFSET: usize = 9;
+
+impl AcpiTableHeader {
+    pub fn new() -> Self {
+        AcpiTableHeader { revision: 2, checksum: 0, oem_id: *b"AKAROS", oem_table_id: *b"ALPHABET", oem_revision: 0, asl_compiler_id: *b"RON ", asl_compiler_revision: 0, ..Default::default() }
```
Rwantare, line 48: Does it make sense to have the oem_id as AKAROS if we're using it?

rminnich: oops, let's pick a better name. This is kind of a nod to the past.

rminnich: try that.

Rwantare (approved):

PR #308, opened 2020-10-20
Add constants for x86

rminnich:


src/arch/x86/x86_64/src/consts/mod.rs
```diff
@@ -0,0 +1,11 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+pub mod msr;
+pub mod x86;
+
+#[allow(non_upper_case_globals)]
```
rminnich, line 5: what's a better place for these names.

Rwantare: Maybe a new crate in Oreboot/src/lib ?
We only have device_tree and print in there

rjoleary: There's also this existing crate: https://docs.rs/kib/4.0.0/kib/

src/arch/x86/x86_64/src/consts/x86.rs
```diff
@@ -0,0 +1,96 @@
...
+ * Paging options in CR3
+ */
+pub const X86_CR3_PWT: u64 = 0x00000008; /* Page Write Through */
+pub const X86_CR3_PCD: u64 = 0x00000010; /* Page Cache Disable */
+pub const X86_CR3_PCID_MASK: u64 = 0x00000fff; /* PCID Mask */
+
+/*
+ * Intel CPU features in CR4
+ */
+pub const X86_CR4_VME: u64 = 0x00000001; /* enable vm86 extensions */
```
rjoleary, line 61: For the documentation, can you use the triple-slash `///`?

src/arch/x86/x86_64/src/consts/mod.rs
rminnich: I'd rather not do a full dependency for a few constants. I'll put it in src/lib

PR #309, opened 2020-10-21
Add Driver::pread_exact method.

doriath:
This is similar to std::io::Read::read_exact, that reads the exact
number of bytes to fill in the input buffers. The current code often
assumes that the buffer will be fully filled in (especially when reading
u32 into a 4 byte buffer), this method will make it much easier to write
it correctly.

### DIFF
--- a/src/drivers/model/src/lib.rs
+++ b/src/drivers/model/src/lib.rs
@@ -1,5 +1,4 @@
 #![no_std]
-#![deny(warnings)]
 
 pub type Result<T> = core::result::Result<T, &'static str>;
 
@@ -20,4 +19,22 @@ pub trait Driver {
     fn pwrite(&mut self, data: &[u8], pos: usize) -> Result<usize>;
     /// Shutdown the device.
     fn shutdown(&mut self);
+    /// Reads the exact number of bytes to fill in the `data`.
+    /// Returns ok if `data` is empty.
+    fn pread_exact(&self, mut data: &mut [u8], mut pos: usize) -> Result<()> {
+        while !data.is_empty() {
+            match self.pread(&mut data, pos) {
+                Ok(0) => return Err("unexpected eof"),
+                Ok(x) => {
+                    data = &mut data[x..];
+                    pos += x;
+                }
+                Err(err) => return Err(err),
+            }
+        }
+        Ok(())
+    }
 }
+
+#[cfg(test)]
+mod tests;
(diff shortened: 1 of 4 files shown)


src/drivers/model/src/lib.rs
```diff
@@ -20,4 +19,16 @@ pub trait Driver {
     fn pwrite(&mut self, data: &[u8], pos: usize) -> Result<usize>;
     /// Shutdown the device.
     fn shutdown(&mut self);
+    /// Reads the exact number of bytes to fill in the `data`.
+    /// Returns ok if `data` is empty.
+    fn pread_exact(&self, data: &mut [u8], pos: usize) -> Result<()> {
+        let mut r = 0;
+        while r < data.len() {
```
fotonick, line 26: Is it possible to read more data than you asked for?

src/drivers/model/src/lib.rs
```diff
@@ -20,4 +19,16 @@ pub trait Driver {
...
+            r += self.pread(&mut data[r..], pos + r)?;
```
fotonick, line 27: Do we want to guard against an infinite loop of reading 0 bytes?

fotonick (commented):
I love mirroring the semantics of something well known for readability.

src/drivers/model/src/lib.rs
doriath: No, unless the pread implementation is invalid.

src/drivers/model/src/lib.rs
doriath: Added protection against reading 0 bytes.

src/drivers/model/src/lib.rs
fotonick: Sounds good. Just wanted you to think about it.

src/drivers/model/src/lib.rs
fotonick: Thanks. Does zero bytes mean EOF? It doesn't mean time out or some other random failure condition (like C libraries like to alias together)? If our spec doesn't specify, one possibility is that we may wish to constrain the return type of `pread` to `Result<NonZeroUsize>` to force future implementers of `Driver` to give everything a well defined meaning. I might be over-thinking this, but this kind of ambiguity does have a habit of breeding bugs.

fotonick (approved):

PR #310, opened 2020-10-21
Remove all unwrap calls from device tree crate.

doriath:
Because we have a proper error handling in the iterator now, we can
return an error instead of panicing.


Rwantare:
Can we use the ['?'](https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/the-question-mark-operator-for-easier-error-handling.html)  operator to propagate the errors here?

Rwantare:
> Can we use the ['?'](https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/the-question-mark-operator-for-easier-error-handling.html) operator to propagate the errors here?

I guess not, if next has to return an Option

Rwantare (approved):

src/lib/device_tree/src/lib.rs
```diff
@@ -185,11 +185,19 @@ pub struct FdtIterator<'a, D: Driver> {
 impl<'a, D: Driver> FdtIterator<'a, D> {
     #[allow(clippy::should_implement_trait)]
     pub fn next<'b>(&'b mut self) -> Option<Result<Entry<'b, SectionReader<'a, SectionReader<'a, D>>>>> {
```
rjoleary, line 187: Would it make the error handling simpler if you swap the return to a `Result<Option<...`? Might not be possible if it has to match a certain signature.

doriath: We do not have to match any signature (we stopped using standard iterator trait here).

I wasn't sure if it will be ergonomic for the users of the iterator if we swap the return type, but it seems it is actually even better than before. Added a new commit to swap the return type.

Rwantare (approved):

PR #311, opened 2020-10-21
Lock version of heapless dependency.

doriath:
Running `make clippy` shows that src/drivers/uart does not specify the
exact version to use and that it will become an error in future
versions. To avoid a future breakage, just lock the version now.

### DIFF
--- a/src/drivers/uart/Cargo.toml
+++ b/src/drivers/uart/Cargo.toml
@@ -11,10 +11,8 @@ opt-level = 'z'  # Optimize for size.
 model = { path = "../model"}
 clock = { path = "../clock"}
 register = "0.3.2"
-
 # Heapless does not build for rv32imc.
-[dependencies.heapless]
-optional = true
+heapless = { version = "0.4.4", optional = true }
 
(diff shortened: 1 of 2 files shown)


Rwantare:
Can we lock to something newer? What does heapless 0.5.6 break?

rminnich (approved):

doriath:
Updated to 0.5.6 as recommended by Titus.

doriath:
It looks like 0.5.6 actually didn't work: https://dev.azure.com/azure0427/Oreboot%20Pipeline/_build/results?buildId=1278&view=logs&j=8ddc64b0-de54-5818-74b0-10cb5a7edecf&t=3c05a52b-1388-530f-3d1c-6ae2f17abbf4

Trying 0.5.4

doriath:
It seems there are some weird dependencies and 0.5.4 also doesn't work. Switching back to 0.4.4

fotonick (approved):

rminnich:
lert's go! commit.

PR #319, opened 2020-10-26
Run clippy on sifive/hifive mainboard.

doriath:
This change creates a separate list for BROKEN crates that do not work
with clippy yet (previuosly we shared the list for `cargo test`). It
also removes sifive/hifive mainboard from broken list and fixes all
existing errors.

### DIFF
--- a/src/mainboard/sifive/hifive/src/main.rs
+++ b/src/mainboard/sifive/hifive/src/main.rs
@@ -13,7 +13,6 @@ use core::{fmt::Write, ptr};
 use device_tree::print_fdt;
 use model::Driver;
 use payloads::payload;
-use print;
 use soc::clock::Clock;
 use soc::ddr::DDR;
 use soc::is_qemu;
--- a/src/soc/sifive/fu540/src/ux00.rs
+++ b/src/soc/sifive/fu540/src/ux00.rs
@@ -37,14 +37,14 @@ pub const PHY_RX_CAL_DQ1_0_OFFSET: u64 = 16;
 // This is a 64-bit machine but all this action seems to be on 32-bit values.
 // No idea why this is.
 // index is a word offset.
-fn poke(pointer: u32, index: u32, value: u32) -> () {
+fn poke(pointer: u32, index: u32, value: u32) {
     let addr = (pointer + (index << 2)) as *mut u32;
     unsafe {
         ptr::write_volatile(addr, value);
     }
 }
 
-fn poke64(pointer: u32, index: u32, value: u64) -> () {
+fn poke64(pointer: u32, index: u32, value: u64) {
     let addr = (pointer + (index << 2)) as *mut u64;
     //let addr1 = (pointer + (index << 2) + 4) as *mut u32;
     unsafe {
@@ -56,12 +56,12 @@ fn poke64(pointer: u32, index: u32, value: u64) -> () {
     }
 }
 
-fn set(pointer: u32, index: u32, value: u32) -> () {
+fn set(pointer: u32, index: u32, value: u32) {
     let v = peek(pointer, index);
     poke(pointer, index, v | value);
 }
 
-fn clr(pointer: u32, index: u32, value: u32) -> () {
+fn clr(pointer: u32, index: u32, value: u32) {
     let v = peek(pointer, index);
     poke(pointer, index, v & value);
 }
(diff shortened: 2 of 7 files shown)


src/mainboard/sifive/hifive/src/main.rs
```diff
@@ -92,40 +91,40 @@ pub extern "C" fn _start_boot_hart(_hart_id: usize, fdt_address: usize) -> ! {
 
     let w = &mut print::WriteTo::new(uart0);
 
-    write!(w, "## ROM Device Tree\r\n").unwrap();
+    writeln!(w, "## ROM Device Tree").unwrap();
```
fotonick, line 94: Usually when I see a `\r`, I pause because it's usually there very deliberately. Do you know why it was there? Will `writeln!` preserve it?

src/soc/sifive/fu540/src/ux00.rs
```diff
@@ -196,7 +196,7 @@ pub fn ux00ddr_phy_fixup() -> u64 {
 
     //let mut fails: u64 = 0;
     let mut slicebase: u32 = 0;
-    let mut dq: u32 = 0;
+    //let mut dq: u32 = 0;
```
fotonick, line 199: Delete rather than comment out.

doriath: I left it as commented because there is other code nearby (e.g. 2 lines above and inside for loop below) that is also commented (and some is even using this dq variable). I would like to suggest to leave it commented as is, and clean up all commented code in some followup PR. WDYT?

src/mainboard/sifive/hifive/src/main.rs
doriath: No idea why it was here, but it seems like the code in this file is randomly using \n and \r\n (e.g. 3 lines below in line 98 it uses just `\n`). Because this is mostly for debugging, and the CI tests pass, I would like to suggest to just drop \r and use writeln everywhere in this file. If at any point in time we find that \r is actually needed, then we can easily change it back and we will have clear reason that we could document at the same time. WDYT?

rminnich: does writeln send a \r or does it assume it? If it does not send \r then that's a problem.

fotonick: "On all platforms, the newline is the LINE FEED character (\n/U+000A) alone (no additional CARRIAGE RETURN (\r/U+000D)." —https://doc.rust-lang.org/std/macro.writeln.html

Ron, can you clarify why `\r` is necessary? It appears that some of the old code had it and some didn't.

rjoleary: Yes `\r\n` is necessary.

`\r` is carriage return; goes to the start of the current line. `\n` is newline; goes down one line without shifting horizontally.

Something like "hello\nworld" would print like this:

```
hello
     world
```

You might wonder why you've never had to use `\r` before.

This is because application-level code in Linux is using a "cooked" TTY. In the cooked mode, `\n` is automatically replaced by `\r\n` before sending out to serial. Different operating systems do this differently (like Windows expects all the files to contain \r\n).

You can experience raw uncooked power in Linux with:

```
~$ stty raw
~$ printf "hello\nworld"
hello
     world~$ 
```

Also, running oreboot in qemu with `-nographic` is misleading because it piggybacks off of the surrounding TTY in Linux. But, I assure you when you run on hardware and connect a serial device, if you only use `\n` the newlines will look weird like in the first example.

fotonick: Thanks, Ryan!

Would it make sense to write our own `writern!` macro that does `\r\n`? Or can we add a custom linting rule to squawk at `writeln!` and `write!` with a bare `\n`? (No idea how one would best add a custom linting rule.)

src/soc/sifive/fu540/src/ux00.rs
fotonick: SGTM

src/mainboard/sifive/hifive/src/main.rs
doriath: Thank you for the detailed explanation Ryan! I reverted the changes to `write!` and disabled the check in clippy. I also created #321 to focus on this (if possible, I would like to separate the simple clippy fixes from fixing the write! macros).

fotonick (approved):

Issue #283, opened 2020-09-18
Build error when building all mainboards


GanShun:
One of the boards is throwing up this error when I build all mainboards:

error[E0658]: `if` is not allowed in a `const`
  --> /home/ganshun/.cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.36/src/int/specialized_div_rem/mod.rs:68:22
   |
68 |   const USE_LZ: bool = if cfg!(target_feature = "thumb-mode") {
   |  ______________________^
69 | |     // ARM thumb targets have CLZ instructions if the instruction set of ARMv6T2 is supported. This
70 | |     // is needed to successfully differentiate between targets like `thumbv8.base` and
71 | |     // `thumbv8.main`.
...  |
77 | |     cfg!(target_feature = "v5te")
78 | | };
   | |_^
   |
   = note: see issue #49146 <https://github.com/rust-lang/rust/issues/49146> for more information
   = help: add `#![feature(const_if_match)]` to the crate attributes to enable

error[E0658]: `if` is not allowed in a `const`
  --> /home/ganshun/.cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.36/src/int/specialized_div_rem/mod.rs:68:22
   |
68 |   const USE_LZ: bool = if cfg!(target_feature = "thumb-mode") {
   |  ______________________^
69 | |     // ARM thumb targets have CLZ instructions if the instruction set of ARMv6T2 is supported. This
70 | |     // is needed to successfully differentiate between targets like `thumbv8.base` and
71 | |     // `thumbv8.main`.
...  |
77 | |     cfg!(target_feature = "v5te")
78 | | };
   | |_^
   |
   = note: see issue #49146 <https://github.com/rust-lang/rust/issues/49146> for more information
   = help: add `#![feature(const_if_match)]` to the crate attributes to enable

fotonick:
I can reproduce. I can also confirm that i'm using the versions that we locked in `make firsttime`:

    cargo --version  = cargo 1.44.0-nightly (8751eb301 2020-04-21)
    rustc --version  = rustc 1.44.0-nightly (b2e36e6c2 2020-04-22)
    rustup --version = rustup 1.22.1 (b01adbbc3 2020-07-08)

It looks to me like we haven't locked everything because during the `cargo build`, I see:

      Downloaded compiler_builtins v0.1.36

[...]

PR #320, opened 2020-10-26
Fix emulation/qemu-armv7 mainboard.

doriath:
This change fixes the rust code (broken by recent refactorings) and
fixes the fixed-dtfs layout to provide enough space for the actual
bootblob (it is over 6M but only 512K was reserved).

Ref #283

### DIFF
--- a/src/mainboard/emulation/qemu-armv7/fixed-dtfs.dts
+++ b/src/mainboard/emulation/qemu-armv7/fixed-dtfs.dts
@@ -13,14 +13,12 @@
             area@0 {
                 description = "Boot Blob";
                 compatible = "ore-bootblob";
-                offset = <0x0>;
-                size = <0x80000>; // 512KiB
+                size = <0x800000>; // 8MiB
                 file = "$(TARGET_DIR)/bootblob.bin";
             };
             area@1 {
                 description = "Fixed DTFS";
                 compatible = "ore-dtfs";
-                offset = <0x80000>;
                 size = <0x80000>; // 512KiB
                 file = "$(TARGET_DIR)/fixed-dtfs.dtb";
             };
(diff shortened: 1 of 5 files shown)


src/mainboard/emulation/qemu-armv7/fixed-dtfs.dts
```diff
@@ -14,56 +14,56 @@
...
                 offset = <0x0>;
-                size = <0x80000>; // 512KiB
+                size = <0x880000>; // 8.5MiB
                 file = "$(TARGET_DIR)/bootblob.bin";
             };
             area@1 {
                 description = "Fixed DTFS";
                 compatible = "ore-dtfs";
-                offset = <0x80000>;
+                offset = <0x880000>;
```
fotonick, line 23: Could omit all of these `offset` entries if you wanted to; IIRC they're auto-computed. I'm not sure that being explicit here actually adds any value to a human reader.

src/mainboard/emulation/qemu-armv7/src/main.rs
```diff
@@ -24,9 +23,10 @@ pub extern "C" fn _start() -> ! {
     uart_driver.pwrite(b"Welcome to oreboot\r\n", 0).unwrap();
     let s = &mut [uart_driver];
     let console = &mut DoD::new(s);
+    let mut w = print::WriteTo::new(console);
+    let spi = SliceReader::new(DTB);
 
-    if let Err(err) = print_fdt(console) {
-        let mut w = print::WriteTo::new(console);
+    if let Err(err) = print_fdt(&spi, &mut w) {
         write!(w, "error: {}\n", err).expect(err);
```
fotonick, line 30: I'm surprised that clippy isn't suggesting `writeln!`.

fotonick (dismissed):
These net negative patches are so satisfying. Thanks for all the good work.

rjoleary (approved):

src/mainboard/emulation/qemu-armv7/fixed-dtfs.dts
doriath: Good point, removed the offset.

src/mainboard/emulation/qemu-armv7/src/main.rs
doriath: Clippy is not enabled on all crates yet. There is pretty long list of BROKEN_CRATES_TO_CLIPPY in the Makefile (I am starting with #319), which is why this was not caught.

fotonick (approved):

PR #322, opened 2020-10-27
Fix clippy errors for majority of crates.

doriath:
The only few crates left are mainboards that are built with custom_targets and xbuild - regular `cargo clippy` does not work for them.


src/vendorcode/fsp/coffeelake/build.rs
```diff
@@ -1,143 +0,0 @@
-extern crate bindgen;
```
rjoleary, line 1: Why is this being removed?

rjoleary: It'll only build after running `git submodule update --init --recursive`. Perhaps we should add this to the CI.

src/mainboard/opentitan/crb/src/main.rs
```diff
@@ -9,35 +9,34 @@ use core::fmt::Write;
...
-    m.pwrite(&[0x1bu8, 0xd7u8, 0x79u8, 0x1fu8], 0x40070018);
-    m.pwrite(&[0x60u8, 0x08u8, 0x00u8, 0x00u8], 0x4007001c);
-    m.pwrite(&[0xc2u8, 0x40u8, 0x14u8, 0x06u8], 0x40070020);
-    m.pwrite(&[0x07u8, 0x92u8, 0x28u8, 0x0bu8], 0x40070024);
-    m.pwrite(&[0x4cu8, 0xe3u8, 0x3cu8, 0x10u8], 0x40070028);
-    m.pwrite(&[0x91u8, 0x34u8, 0x51u8, 0x15u8], 0x4007002c);
-    m.pwrite(&[0xd6u8, 0x85u8, 0x65u8, 0x1au8], 0x40070030);
-    m.pwrite(&[0x1bu8, 0xd7u8, 0x79u8, 0x1fu8], 0x40070034);
-    m.pwrite(&[0x60u8, 0x08u8, 0x00u8, 0x00u8], 0x40070038);
+    m.pwrite(&[0xc2u8, 0x40u8, 0x14u8, 0x06u8], 0x40070004).unwrap();
```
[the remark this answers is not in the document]
doriath: Good catch, I was not expecting the size increase by this change. Changed it to explicit ignore the errors and added a comment to explain why we are ignoring them.

src/vendorcode/fsp/coffeelake/build.rs
doriath: Reverted this change and added it back to the list of crates that do not work with clippy yet.

rjoleary (approved):

PR #323, opened 2020-10-28
Upstreaming romecrb: add msrs.rs

Rwantare:


src/mainboard/amd/romecrb/src/msr.rs
```diff
@@ -0,0 +1,428 @@
+use core::fmt::Write;
+use x86_64::instructions::{rdmsr, wrmsr};
+/// Only read the MSR
```
rjoleary, line 3: Please put a blank line before this line.

rjoleary (commented):
Can you squash the commits?
Otherwise, everything else looks great!

src/mainboard/amd/romecrb/src/msr.rs
rminnich: should we follow the go style of doing this: 
/// read_msr reads an msr, compares it to a "known good" value, and prints a message, indicating
if matches or is different from that value.

src/mainboard/amd/romecrb/src/msr.rs
```diff
@@ -0,0 +1,428 @@
...
+    let checked_value = rdmsr(address);
+    writeln!(w, " -- wrmsr: and got {:x}; \r", checked_value).expect("Failed to write!");
+}
+
+/// To be used on Write-Only MSRs
+fn _write_msr(w: &mut print::WriteTo, address: u32, value: u64) {
+    unsafe { wrmsr(address, value) }
+}
+
+pub fn msrs(w: &mut print::WriteTo) {
```
rminnich, line 28: might as well document this one a little bit too.

rminnich (approved):

PR #324, opened 2020-10-29
Upstreaming romecrb to master

Rwantare:
Trying to clean up the romecrb code so that we don't get any further away from master.


src/mainboard/amd/romecrb/fixed-dtfs.dts
```diff
@@ -13,43 +13,43 @@
             area@0 {
                 description = "Boot Blob";
                 compatible = "ore-bootblob";
-                offset = <0x3f0000>;
-                size = <0x10000>;
+                offset = <0x3e0000>;
```
rminnich, line 16: was this changed for your board? I will need to test these numbers on CRB I guess?

rminnich (approved):
let's go for it.

src/mainboard/amd/romecrb/fixed-dtfs.dts
Rwantare: This matches your fork

PR #326, opened 2020-11-03
Add the port 80 uart to the rome

rminnich:
The port 80 uart is very useful when the serial is dead for some reason.


rminnich:
I realize this is not The Best Code Ever but I hope the intent is clear. 

This is actually something we are using NOW to debug a problem.

Rwantare (commented):
Should DIT be tracked on master? It only works on your machine

rminnich:
removed DIT

Rwantare (approved):
Looks good to me

rjoleary (approved):

rminnich:
it was pretty cool to see the console output on the bmc port 80 display ...

PR #327, opened 2020-11-17
add P0 uart

rminnich:
Signed-off-by: Ronald G Minnich <rminnich@gmail.com>

### DIFF
--- a/src/drivers/uart/src/amdmmio.rs
+++ b/src/drivers/uart/src/amdmmio.rs
@@ -1,0 +1,131 @@
+/*
+ * This file is part of the oreboot project.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ */
+
+use core::ops;
+use model::*;
+
+use register::mmio::{ReadOnly, ReadWrite};
+use register::register_bitfields;
+
+const RETRY_COUNT: u32 = 100_000;
+const COM1: usize = 0xfedc9000;
+
+// We fill out as little of this as possible.
+// We're firmware and should never plan to use it
+// with all its features.
+#[repr(C)]
+pub struct RegisterBlock {
+    // pathetically, this is also the DLL.
+    d: ReadWrite<u32, D::Register>, /* data register */
+    // This is the IER when we are not in DLAB mode.
+    // We never intend to set inerrupts, so we just set
+    // it up as thing we write to.
+    dlm: ReadWrite<u32, D::Register>, /* data register */
+    _8: u32,
+    lcr: ReadWrite<u32, LCR::Register>,
+    _10: u32,
+    stat: ReadOnly<u32, STAT::Register>, /* status */
+}
+
+pub struct AMDMMIO {
+    base: usize,
+    // TODO: implement baudrate
+    //baudrate: u32,
+}
+
+impl ops::Deref for AMDMMIO {
+    type Target = RegisterBlock;
+    fn deref(&self) -> &Self::Target {
+        unsafe { &*self.ptr() }
+    }
+}
+
+register_bitfields! {
+    u32,
+    D [
+        Data OFFSET(0) NUMBITS(8) []
+    ],
+    STAT[
+        DR OFFSET(0) NUMBITS(1) [],
+        THRE OFFSET(6) NUMBITS(1) [],
+        TDRE OFFSET(5) NUMBITS(1) []
+    ],
+    FCR[
+        FIFOENABLE OFFSET(0) NUMBITS(1) []
+    ],
+    LCR[
+        BITSPARITY OFFSET(0) NUMBITS(3) [
+            EIGHTN1 = 3
+        ],
+        DLAB OFFSET(7) NUMBITS(1)[
+            Data = 0,
+            BaudRate = 1
+        ]
+    ]
+}
+
+impl AMDMMIO {
+    pub fn new(base: usize) -> AMDMMIO {
+        AMDMMIO { base: base }
+    }
+
+    pub fn com1() -> AMDMMIO {
+        AMDMMIO { base: COM1 }
+    }
+
+    /// Returns a pointer to the register block
+    fn ptr(&self) -> *const RegisterBlock {
+        self.base as *const _
+    }
+}
+
+impl Driver for AMDMMIO {
+    fn init(&mut self) -> Result<()> {
+        self.lcr.write(LCR::BITSPARITY::EIGHTN1 + LCR::DLAB::BaudRate);
+        self.d.set(1);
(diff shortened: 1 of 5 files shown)


src/drivers/uart/src/lib.rs
```diff
@@ -17,3 +17,11 @@ pub mod opentitan;
 pub mod pl011;
 #[cfg(feature = "sifive")]
 pub mod sifive;
+#[cfg(feature = "amdmmio")]
+pub mod amdmmio;
+
+/* Calculate divisor. Do not floor but round to nearest integer. */
+pub fn uart_baudrate_divisor(baudrate: usize, refclk: usize, oversample: usize) -> usize {
```
fotonick, line 24: Not called anywhere?

src/mainboard/amd/romecrb/src/main.rs
```diff
@@ -201,7 +215,10 @@ pub extern "C" fn _start(fdt_address: usize) -> ! {
     uart0.pwrite(b"Welcome to oreboot\r\n", 0).unwrap();
     debug.init().unwrap();
     debug.pwrite(b"Welcome to oreboot - debug port 80\r\n", 0).unwrap();
-    let s = &mut [debug as &mut dyn Driver, uart0 as &mut dyn Driver];
+    let p0 = &mut AMDMMIO::new(0xfedc9000);
```
fotonick, line 218: Comment magic number or set as a const at the top?

fotonick (commented):
Is any of this unit-testable or not really?

Rwantare:
You can run ```make format``` to get past the CheckQuality error

src/mainboard/amd/romecrb/src/main.rs
```diff
@@ -201,7 +215,10 @@ pub extern "C" fn _start(fdt_address: usize) -> ! {
...
+    p0.init().unwrap();
+    p0.pwrite(b"Welcome to oreboot - debug port 80\r\n", 0).unwrap();
+    let s = &mut [debug as &mut dyn Driver, uart0 as &mut dyn Driver, p0 as &mut dyn Driver, ];
```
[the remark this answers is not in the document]
rminnich: Fixed. Clippy did not find it.

src/drivers/uart/src/amdmmio.rs
```diff
@@ -0,0 +1,136 @@
...
+ * GNU General Public License for more details.
+ */
+
+use core::ops;
+use model::*;
+
+use register::mmio::{ReadOnly, ReadWrite};
+use register::register_bitfields;
+
+const RETRY_COUNT: u32 = 100_000;
```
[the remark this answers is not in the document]
rminnich: This is a pretty common retry count for uarts, which is where I got it. It is useful because it is so large that it makes it detectable that you are always timing out when you watch uart output. Were you to make it much closer to a reasonable value you might miss that you're not testing the right bits in the register, or that they are not operating correctly. 

Not making this up: it's caught this mistake before for me and it caught it recently in our i8250 driver where we were in fact always timing out and not checking the right bits. The performance is so bad it stands out. You don't want timeout performance to be good.

src/drivers/uart/src/amdmmio.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// Returns a pointer to the register block
+    fn ptr(&self) -> *const RegisterBlock {
+        self.base as *const _
+    }
+}
+
+impl Driver for AMDMMIO {
+    fn init(&mut self) -> Result<()> {
+        // This does not work.
+        if false {
```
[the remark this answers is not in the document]
rminnich: because Titus it going to tell me how to fix it when he reviews it :-)

src/drivers/uart/src/lib.rs
rminnich: yet. :-)

src/mainboard/amd/romecrb/src/main.rs
rminnich: I am halfway thinking of allowing an Option parameter and if it is not passed having that default to COM1 as known to the driver, wdyt? 99.99999% of the uses will be for COM1, so why make them have to pass the parameter.

rminnich: better yet: AMDMMIO:com1()
then it's clear what's up

src/drivers/uart/src/amdmmio.rs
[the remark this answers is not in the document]
rminnich: it won't go in until this is fixed, not to worry.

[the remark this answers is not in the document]
Rwantare: Try two separate modifications to LCR. write() sets all the other fields to zero, so you could try modify(). Then you don't have to use the + within, which I don't think I have seen used.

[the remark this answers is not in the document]
rminnich: well it is supposed to work, it's in the tock docs, but I'll try it. 

One of the good things about the way I do it is that you want to set all the bits in one write, sometimes the read-modify-write won't work in some cases (some bits can be writeonly). But I'll try it friday!

src/drivers/uart/src/amdmmio.rs
```diff
@@ -0,0 +1,140 @@
...
+
+impl Driver for AMDMMIO {
+    fn init(&mut self) -> Result<()> {
+        // This does not work.
+//         if false {
+            self.lcr.modify(LCR::BITSPARITY::EIGHTN1);
+            self.lcr.modify(LCR::DLAB::SET);
+            self.d.set(1);
+            self.dlm.set(0);
+            self.lcr.modify(LCR::DLAB::CLEAR);
```
rminnich, line 98: this does not seem to be defined, right? I will defined it or something like it.

Rwantare: "SET" and "CLEAR" are defined for all 1 bit fields

rminnich: the name should not be set and clear anyway, more like BaudDivisor and DataMode or some such?

Rwantare: Since DLAB is Divisor Latch Access Bit, you would want to name it in a way that signifies that access to the divisor latch is being enabled or disabled. In which case Enable and Disable are more fitting, assuming the reader knows what DLAB is.

rminnich: e.g.:
let control: InMemoryRegister<u32, Control::Register> = InMemoryRegister::new(0)
control.write(Contol::BYTE_COUNT.val(0) +
              Contol::ENABLE::Yes +
              Contol::LENGTH.val(10));

Rwantare: I'd forgotten about that. Don't we want to use the 'InMemoryRegister' type as well?

src/drivers/uart/src/amdmmio.rs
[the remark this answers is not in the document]
rminnich: fixed. I had the OFFSET for the 8n1 wrong, 1 instead of 0. objdump is your friend.

src/drivers/uart/src/amdmmio.rs
rminnich: I think the RegisterBlock gets us that?

Rwantare (approved):

PR #332, opened 2020-11-30
firsttime: add add llvm-tools-preview

rminnich:
This makes sure we have rust-objcopy

Signed-off-by: Ronald G Minnich <rminnich@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,7 @@ firsttime:
 	rustup target add armv7r-none-eabi
 	cargo install $(if $(XBUILD_VER),--version $(XBUILD_VER),) cargo-xbuild
 	cargo install $(if $(BINUTILS_VER),--version $(BINUTILS_VER),) cargo-binutils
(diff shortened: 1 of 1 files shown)


fotonick (changes requested):
Ron, this is already present as part of line 34. I don't believe your change is necessary.

Makefile
```diff
@@ -37,6 +37,7 @@ firsttime:
 	rustup target add armv7r-none-eabi
 	cargo install $(if $(XBUILD_VER),--version $(XBUILD_VER),) cargo-xbuild
 	cargo install $(if $(BINUTILS_VER),--version $(BINUTILS_VER),) cargo-binutils
+	rustup component add llvm-tools-preview
```
rjoleary, line 40: What about the llvm-tools-preview on line 34? Do you think the order matters?

rminnich:
did not fix anything anyway

PR #334, opened 2020-12-03
mainboard/opentitan: Get oreboot booting in QEMU

alistair23:
With this patch oreboot now boots after the OpenTitan boot ROM in QEMU.

### DIFF
--- a/src/mainboard/opentitan/crb/link.ld
+++ b/src/mainboard/opentitan/crb/link.ld
@@ -30,6 +30,10 @@ SECTIONS
     . = 0x20000000;
     .bootblock :
     {
+        /* Ensure that the flash header matches what the OpenTitain ROM expects.
+         */
+        LONG(_stext);
+        _stext = .;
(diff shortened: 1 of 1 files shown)


orangecms:
Looks like a typo snuck in, should it be `OpenTitan` as in https://opentitan.org/?

alistair23:
Fixed!

rminnich (approved):

src/mainboard/opentitan/crb/link.ld
```diff
@@ -30,6 +30,10 @@ SECTIONS
     . = 0x20000000;
     .bootblock :
     {
+        /* Ensure that the flash header matches what the OpenTitain ROM expects.
```
rminnich, line 33: oh damn. OpenTitan not OpenTitain

alistair23: Argh! Fixed here: https://github.com/oreboot/oreboot/pull/336

PR #337, opened 2020-12-06
mainboard/emulation/qemu-aarch64: Get oreboot booting in QEMU

ichihlu-g:
Booting in QEMU aarch64 platform with oreboot as bios then jump to the loaded Linux kernel with processor mode starts from non-secure state.


src/arch/arm64/armv8/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+    fn shutdown(&mut self) {}
+}
+
+// TODO
+// extern "C" {
+//     fn mmu_get() -> usize;
+//    fn mmu_set(i: usize);
+// }
+
+// global_asm!(include_str!("mmu.S"));
```
alistair23, line 46: I always think it's better to just remove commented out code instead of leaving it to bit rot.

src/mainboard/emulation/qemu-aarch64/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+    write!(w, "7").expect("blame ryan");
+    romstage::romstage(&mut w)
+}
+use core::panic::PanicInfo;
+
+pub fn halt() -> ! {
+    loop {
+        // Bug with LLVM marks empty loops as undefined behaviour.
+        // See: https://github.com/rust-lang/rust/issues/28728
+        unsafe { llvm_asm!("" :::: "volatile") }
```
alistair23, line 72: Why not put a `nop` in here?

alistair23 (changes requested):

src/arch/arm64/armv8/src/lib.rs
ichihlu-g: Good suggestion! I removed these lines.

alistair23 (approved):

src/mainboard/emulation/qemu-aarch64/src/main.rs
ichihlu-g: Thanks for pointing that out. I changed this by putting WFE (wait for event) instruction here.

PR #354, opened 2021-01-15
Add command to print stack size of each function

rjoleary:
Run with `make stack-size`.


fotonick:
This is exciting and looks like it matches the description in the [unstable Rust book](https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/emit-stack-sizes.html). Do we have a `make test` target that we might upgrade to exercise this and at least verify that the `.stack_sizes` section in the final binary is present and non-zero size?

azure-pipelines.yml
```diff
@@ -47,6 +47,7 @@ jobs:
       make ciprepare
       cd src/mainboard/emulation/qemu-q35
       make
+      make stack-sizes 2>/dev/null | grep '0x[0-9a-f]\+\s\+[0-9]\+\s\+_start'
```
fotonick, line 50: || echo "Error: Didn't see stack size of _start. make stack-sizes might be broken."

azure-pipelines.yml
```diff
@@ -98,6 +100,7 @@ jobs:
   - script: |
       pushd src/mainboard/emulation/qemu-riscv
```
rminnich, line 101: instead of pushd/popd could you put this in ()
(cd bla && make && make blah blah)
?

orangecms (approved):

rminnich (approved):

PR #364, opened 2021-02-03
upgrade gcc-5 to gcc-6

rminnich:
azure can no longer find gcc-5


rminnich:
This will unblock things

alistair23 (approved):

PR #369, opened 2021-02-06
Factor out toolchain overrides into rust-toolchain file

### LINKED ISSUE #275: Pin versions using toolchain-file
Right now we pin our toolchains, release channel, versions, and specify what components to install by hard-coding several rustup invocations in the `firsttime` make target. I believe that better would be to use the new TOML-formatted `toolchain-file` that's supported in https://github.com/rust-lang/rustup/pull/2438. Unfortunately, it's not in the latest release (1.22.1 from July 8), so we'll have to wait a little while.

An example `toolchain-file` we might use is

    [toolchain]
    channel = "nightly-2020-04-23"
    components = [ "rust-src", "llvm-tools-preview" ]
    targets = [
    "riscv64imac-unknown-none-elf",
    "riscv32imc-unknown-none-elf",
    "armv7r-none-eabi",
    ]

I like that everything Rust is in one place that we can invoke independently of the rest of `make firsttime`.

fotonick:
The `rust-toolchain` file appears to be the official way to check toolchain overrides into version control. https://rust-lang.github.io/rustup/overrides.html?highlight=toml#the-toolchain-file

It appeals to me because it separates data from code.

Fixes #275.

Signed-off-by: Nickolas Fotopoulos <foton@google.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ BROKEN := \
 
 MAINBOARDS := $(filter-out $(BROKEN), $(wildcard src/mainboard/*/*/Makefile))
 
-TOOLCHAIN_VER := nightly-2020-10-25
+TOOLCHAIN_VER := $(shell grep channel rust-toolchain | grep -e '".*"' -o)
 BINUTILS_VER := 0.3.2
 
 .PHONY: mainboards $(MAINBOARDS)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -29,20 +29,15 @@ $(MAINBOARDS):
...
 
 firsttime:
-	rustup override set $(TOOLCHAIN_VER)
-	rustup component add rust-src llvm-tools-preview rustfmt clippy
-	rustup target add riscv64imac-unknown-none-elf
-	rustup target add riscv32imc-unknown-none-elf
-	rustup target add armv7r-none-eabi
-	rustup target add aarch64-unknown-none-softfloat
+	# --default-toolchain is purely an optimization to avoid downloading stable Rust first.
+	curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain $(TOOLCHAIN_VER)
```
orangecms, line 33: I'd put this curl-pipe-sh in `debiansysprepare` to avoid running it for people who would already have rustup installed through other means, since it's available as a package usually (e.g. in Arch) and redundant copies may cause issues.

orangecms (approved):

PR #373, opened 2021-02-16
romecrb: Use vcell for registers (instead of integers).

daym:
Signed-off-by: Danny Milosavljevic <danny.milo@datacom.wien>


daym:
This one is better than https://github.com/oreboot/oreboot/pull/372 --but needs an extra dependency

rminnich (approved):

PR #374, opened 2021-02-20
hack the smn -- check values, set if don't match

rminnich:

### DIFF
--- a/src/mainboard/amd/romecrb/src/main.rs
+++ b/src/mainboard/amd/romecrb/src/main.rs
@@ -6,7 +6,6 @@
 
 use arch::bzimage::BzImage;
 use arch::ioport::IOPort;
-use soc::{smn_write,smn_read};
 use core::fmt::Write;
 use core::panic::PanicInfo;
 use cpu::model::amd_family_id;
@@ -15,6 +14,7 @@ use model::Driver;
 use print;
 use raw_cpuid::CpuId;
 use soc::soc_init;
+use soc::{smn_read, smn_write};
 use uart::amdmmio::AMDMMIO;
 use uart::debug_port::DebugPort;
 use uart::i8250::I8250;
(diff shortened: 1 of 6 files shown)


daym (dismissed):

src/mainboard/amd/romecrb/src/main.rs
```diff
@@ -31,6 +31,26 @@ use core::ptr;
...
+    // the smn device is at (0)
+    unsafe {
+        outl(0xcf8, 0x8000_00b8);
+        outl(0xcfc, a);
+        outl(0xcf8, 0x8000_00bc);
+        inl(0xcfc)
+    }
+}
+
+fn smn_write(a: u32, v: u32) {
```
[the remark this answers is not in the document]
rminnich: these functions should probably be in SOC anyway. More comments coming on your patch to explain why.

src/soc/amd/rome/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+        inl(0xcfc)
+}
+
+unsafe fn smn_write(a: u32, v: u32) {
+        outl(0xcf8, 0x800000b8);
+        outl(0xcfc, a);
+        outl(0xcf8, 0x800000bc);
+        outl(0xcfc, v);
+}
+
```
Rwantare, line 17: Include:
```
 /// Write 32 bits to port
 unsafe fn outl(port: u16, val: u32) {
     llvm_asm!("outl %eax, %dx" :: "{dx}"(port), "{al}"(val));
 }

 /// Read 32 bits from port
 unsafe fn inl(port: u16) -> u32 {
     let ret: u32;
     llvm_asm!("inl %dx, %eax" : "={ax}"(ret) : "{dx}"(port) :: "volatile");
     return ret;
 }
```

Rwantare: I think they're commented out in IOPort

daym (dismissed):

daym (approved):

PR #377, opened 2021-02-24
Add AMD HSMP accessors

daym:
This adds AMD Host System Management Port accessors--and tests them on boot.

It also adds a PCI abstraction and a CPUID check (because it needs both of them).


Rwantare:
You can save some time by running `make --keep-going clippy` before committing, but clippy seems to be upset about new things this week.

Rwantare (changes requested):
Can pci2 just be pci.rs? then I can merge the rest of the pci code into it

daym:
> Can pci2 just be pci.rs? then I can merge the rest of the pci code into it

Sure, I've renamed it now.

src/mainboard/amd/romecrb/src/main.rs
```diff
@@ -200,6 +203,104 @@ fn consdebug(w: &mut impl core::fmt::Write) -> () {
...
+        (extended_model_id << 4) | base_model_id
+    }
+}
+
+fn rome_ff_init(w: &mut impl core::fmt::Write) -> Result<(), &'static str> {
+    let hsmp = HSMP::new(0);
+    unsafe {
+        match hsmp.test(42) {
+            Ok(v) => {
+                write!(w, "HSMP test(42) result: {}\r\n", v);
```
Rwantare, line 234: You could stick .unwrap() at the end of these. I know the code has a lot of warnings already.🤷🏾‍♂️

src/mainboard/amd/romecrb/src/main.rs
```diff
@@ -262,6 +363,9 @@ pub extern "C" fn _start(fdt_address: usize) -> ! {
         msrs(w);
     }
     p[0] = p[0] + 1;
+
+    amd_init(w);
```
Rwantare, line 367: Unwrap here too.

Rwantare (approved):

src/arch/x86/x86_64/src/amdromehsmp.rs
```diff
@@ -0,0 +1,131 @@
...
+// trait: pub unsafe fn test(&self, v: u32) -> Result<u32, u32>
+
+impl HSMP<'_> {
+    pub fn new(nbio: usize) -> Self {
+        let bus = match nbio {
+            0 => 0,
+            1 => 0x20,
+            2 => 0x40,
+            3 => 0x60,
+            _ => panic!("invalid NBIO"),
```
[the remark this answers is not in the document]
daym: Hmm, have to read up some more. I only have one socket populated, so not sure.

Also, hardcoding those PCI bus addresses only works because Ron's to-be-merged patch sets them up that way.

src/mainboard/amd/romecrb/src/main.rs
```diff
@@ -200,6 +203,104 @@ fn consdebug(w: &mut impl core::fmt::Write) -> () {
...
+        Some(vendor) => {
+            if vendor.as_string() != "AuthenticAMD" {
+                panic!("Only AMD is supported");
+            }
+        }
+        None => {
+            panic!("Could not determine whether or not CPU is AMD");
+        }
+    }
+    // write!(w, "CPU Model is: {}\r\n", cpuid.get_extended_function_info().as_ref().map_or_else(|| "n/a", |extfuninfo| extfuninfo.processor_brand_string().unwrap_or("unreadable"),)); // "AMD EPYC TITUS N-Core Processor"
```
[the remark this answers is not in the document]
daym: Heh--I was wondering too, but that's really what it says.

src/mainboard/amd/romecrb/src/main.rs
daym: If it's OK I'll keep those for the time being because I think this thing probably will move to mainboard.rs anyway. I'll bring it up the move tomorrow or Friday (main.rs is already very cluttered with weird specific stuff. Not sure that that's good. On the other hand, being able to `write` is nice)

daym: Even better would be to actually process the error message here. (But rust always mags at me that I should handle the Err case even when I do handle it using `match`. I'll try again...)

src/mainboard/amd/romecrb/src/main.rs
daym: Done.

src/mainboard/amd/romecrb/src/main.rs
daym: I just `write` the detailed message inside the function `amd_init` now--works fine.

src/arch/x86/x86_64/src/amdromehsmp.rs
```diff
@@ -0,0 +1,132 @@
+#![allow(non_snake_case)]
```
rminnich, line 1: This hiearchy of cpu support in oreboot follows coreboot. 
SOC
CPU
ARCH

the soc in an instance of a cpu; soc has extra IP cores. The SMN code should go in soc I suspect, since it's about a given chip and chiplets.

the CPU is an instance of an ARCH. This file should go there I bet?

We've done a poor job on amd so far of honoring these rules. We did a good job on sifive, so let's look there.

ARCH
arch/riscv/rv32/src/lib.rs arch/riscv/rv32/src/mod.rs arch/riscv/rv32/src/reg.rs arch/riscv/rv64/src/lib.rs arch/riscv/rv64/src/mod.rs arch/riscv/rv64/src/reg.rs
note there are no Cargo.toml here


CPU

In cpu we have as below; here we see some Cargo.toml, but no riscv. There's nothing special about riscv cpus yet? 

cpu/armltd/cortex-a53/Cargo.lock cpu/armltd/cortex-a53/Cargo.toml cpu/armltd/cortex-a53/src cpu/armltd/cortex-a9/Cargo.lock cpu/armltd/cortex-a9/Cargo.toml cpu/armltd/cortex-a9/src cpu/lowrisc/ibex/Cargo.lock cpu/lowrisc/ibex/Cargo.toml cpu/lowrisc/ibex/src

SOC
soc/opentitan/earlgrey/Cargo.lock soc/opentitan/earlgrey/Cargo.toml soc/opentitan/earlgrey/src soc/sifive/fu540/Cargo.lock soc/sifive/fu540/Cargo.toml soc/sifive/fu540/src

for the rest of this patch I'l specular on where things ought to be.

For starters, though, I think this file should be in 
soc/amd/rom/src/hsmp.rs

note that, due to namespacing, we don't need to repeat the path name in the file name. hsmp.rs is fine.

src/arch/x86/x86_64/src/lib.rs
```diff
@@ -6,9 +6,11 @@
 
 const PAGE_SIZE: usize = 4096;
 pub mod acpi;
+pub mod amdromehsmp;
 pub mod bzimage;
```
rminnich, line 10: The other thing we learned the hard way: 
arch is the root of the tree 
cpu pulls mods from arch
soc pulls mods from cpu
mainboard should for the most part import soc

anything else turns into a huge mess, as it did in tock, with everything reference everything.
So if we move amdromehsmp to soc/amd/rome/src/hsmp, this pub mod line should not be here.

src/arch/x86/x86_64/src/lib.rs
```diff
@@ -6,9 +6,11 @@
...
 pub mod consts;
 pub mod ioport;
+pub mod pci;
```
rminnich, line 13: probably same for this. If pci is here it's a good bet our structure needs to be reworked.

src/arch/x86/x86_64/src/pci.rs
```diff
@@ -0,0 +1,37 @@
...
+
+pub struct PciAddress {
+    pub segment: u8,  // 4 bits
+    pub bus: u8,      // 8 bits
+    pub device: u8,   // 5 bits
+    pub function: u8, // 3 bits
+    pub offset: u16,  // 12 bits
+}
+
+impl PciAddress {
```
rminnich, line 13: is pci properly part of the architecture? Expecially this register below is more part of the cpu or the soc.

src/mainboard/amd/romecrb/src/main.rs
```diff
@@ -4,12 +4,15 @@
 #![no_main]
 #![feature(global_asm)]
 
+use arch::amdromehsmp::HSMP;
```
rminnich, line 7: use soc::amd::HSMP;

src/mainboard/amd/romecrb/src/main.rs
```diff
@@ -200,6 +203,110 @@ fn consdebug(w: &mut impl core::fmt::Write) -> () {
 }
 //global_asm!(include_str!("init.S"));
 
+// https://github.com/gz/rust-cpuid/issues/26#issuecomment-785280143
```
rminnich, line 206: These model functions probably needs to be in cpu::amd::something-or-other.

src/mainboard/amd/romecrb/src/main.rs
```diff
@@ -200,6 +203,110 @@ fn consdebug(w: &mut impl core::fmt::Write) -> () {
...
+    let base_model_id = s.model_id();
+    let extended_model_id = s.extended_model_id();
+    if base_family_id < 0xF {
+        base_model_id
+    } else {
+        (extended_model_id << 4) | base_model_id
+    }
+}
+
+fn rome_ff_init(w: &mut impl core::fmt::Write) -> Result<(), &'static str> {
```
rminnich, line 229: This would then go in soc::amd::whatever and we would call it 
soc_init I think.

src/mainboard/amd/romecrb/src/main.rs
```diff
@@ -200,6 +203,110 @@ fn consdebug(w: &mut impl core::fmt::Write) -> () {
...
+            }
+            Err(e) => {
+                write!(w, "HSMP interface version error: {}\r\n", e).unwrap();
+            }
+        }
+    }
+    Ok(())
+}
+
+fn amd_init(w: &mut impl core::fmt::Write) -> Result<(), &str> {
```
rminnich, line 260: and since we're using namespace scoping, and this is a cpu-level function, maybe
cpu_init

src/mainboard/amd/romecrb/src/main.rs
```diff
@@ -262,6 +369,14 @@ pub extern "C" fn _start(fdt_address: usize) -> ! {
         msrs(w);
     }
     p[0] = p[0] + 1;
+
+    match amd_init(w) {
```
rminnich, line 373: cpu_init

rminnich (commented):
Had some comments, it will be easier if we get the organization right now rather than go back and fix it, if that's ok. Tock turned into a huge mess because they did not get this right from the start.

rminnich (dismissed):

src/arch/x86/x86_64/src/lib.rs
daym: Moved to soc rome.

src/mainboard/amd/romecrb/src/main.rs
daym: Done

src/mainboard/amd/romecrb/src/main.rs
daym: Done.

src/mainboard/amd/romecrb/src/main.rs
daym: Done

src/mainboard/amd/romecrb/src/main.rs
daym: Done

rminnich (dismissed):

src/arch/x86/x86_64/src/pci.rs
daym: Moved to SOC (the conservative choice).

src/mainboard/amd/romecrb/src/main.rs
daym: Done

Rwantare (approved):

PR #380, opened 2021-02-27
asrock/a300m-stx: run init() without UART1

orangecms:


orangecms:
Running the first init bits works fine thus far. Next, I'm going to gradually copy over the cpu init etc from romecrb. :)

rminnich (approved):

PR #381, opened 2021-03-01
Introduce HPET Timer and slow down port 0x80 output by using it

daym:
This introduces a HPET Timer and slow down port 0x80 output by using it.

Note that HPET has to be enabled, otherwise it will hang on the first sleep (for example on the first write to `debug_port`).


src/drivers/uart/src/debug_port.rs
```diff
@@ -29,7 +29,10 @@ impl<'a> Driver for DebugPort<'a> {
         for (_i, &c) in data.iter().enumerate() {
             let mut s = [0u8; 1];
             s[0] = c;
-            self.timer.sleep(4_000_000); // that's in fs
+            // 1 microsecond
+            for _j in 1..125 {
```
daym, line 33: _j should be 0..125, otherwise the length is 124.

Also, isn't it 0.96 us as it is now?

orangecms: Oh well, I will start at `0`, right. :) And yes, that makes it half a microsecond. Whoops.

I'll drop a comment regarding 32-bit support.

orangecms (dismissed):
I've fixed it all up, so clippy is happy again, and increased to .5us. Now the output is readable :)

Thank you so much! \o/

orangecms (approved):

PR #390, opened 2021-03-08
asrock/a300m-stx: ACPI bootstrapping

orangecms:
This is as far as it gets, and it allows for booting into a u-root shell
with the kernel from
https://github.com/oreboot/binaries/tree/master/rminnich/romecrb


daym (dismissed):

orangecms:
Fur further iteratoins, see boot log from here and original `lspci` logs from the stock firmware:
https://gist.github.com/orangecms/428a617d4081299837ddf87a44f3e106

rminnich (approved):

PR #397, opened 2021-03-18
DF: Add ios_secondary_bus to FabricComponent.

daym:
This adds minimal information we need to be able to get PCIe to work. In this case, that's the secondary bus for the IOS of an IOMS.


orangecms (approved):
boots on A300 + Picasso

PR #398, opened 2021-03-26
code to set up the data fabric registers

rminnich:


daym (approved):
Looks good to me.

As future work, let's work on reducing those changes to the ones that are really necessary. Right now I think this is the right approach.

PR #401, opened 2021-03-29
Introduce common MP mailbox interface--and use it

daym:
Since we have multiple very similar MP mailbox interfaces now, let's introduce a common mailbox interface accessor for all of them. 

Since we know Picasso MP1 from coreboot, publish that fact.


src/soc/amd/rome/src/lib.rs
```diff
@@ -3,30 +3,32 @@
 
 pub mod hsmp;
 
+use hsmp::HSMP;
+
 pub fn soc_init(w: &mut impl core::fmt::Write) -> Result<(), &'static str> {
```
rminnich, line 8: I notice there is no error return. Should there be?

rminnich (dismissed):

src/soc/amd/rome/src/lib.rs
daym: Eventually yes--but right now, we are not using HSMP for anything. So it doesn't make sense to stop the entire boot process (for example) just because it returns something weird.

rminnich (approved):

rminnich (approved):

PR #402, opened 2021-03-30
mpmailbox: Clean up source code.

daym:
Tiny clean up.

### DIFF
--- a/src/soc/amd/common/mp/src/mpmailbox.rs
+++ b/src/soc/amd/common/mp/src/mpmailbox.rs
@@ -13,19 +13,19 @@ pub type Result<T> = core::result::Result<T, MPMailboxCallError>;
 pub struct MPMailbox<const MESSAGE_ARGUMENT_COUNT: usize> {
     message_id_smn_address: u32,
     message_response_smn_address: u32,
-    message_first_arguments_smn_address: u32,
+    message_first_argument_smn_address: u32,
 }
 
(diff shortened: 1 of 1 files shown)


src/soc/amd/common/mp/src/mpmailbox.rs
```diff
@@ -13,19 +13,19 @@ pub type Result<T> = core::result::Result<T, MPMailboxCallError>;
 pub struct MPMailbox<const MESSAGE_ARGUMENT_COUNT: usize> {
     message_id_smn_address: u32,
```
rminnich, line 14: these identifiers feel a little C++-like to me. Personally I like it shorter:
cmd
status
args

and then I think it'd be nice if we had an smn_address type.

rminnich (approved):

src/soc/amd/common/mp/src/mpmailbox.rs
daym: >these identifiers feel a little C++-like to me. Personally I like it shorter:
>cmd
>status
>args

I prefer long names since it's more readable and intellisense exists anyway--but no strong opinion on it.

>then I think it'd be nice if we had an smn_address type.

I agree. I've wanted to do it before, but didn't want to get all architecture astronaut on oreboot yet :)

See https://doc.rust-lang.org/rust-by-example/generics/new_types.html for how to do it.

PR #404, opened 2021-04-02
soc/rome:Have the rome SOC struct contain topology and HSMP

rminnich:

### DIFF
--- a/src/soc/amd/rome/src/lib.rs
+++ b/src/soc/amd/rome/src/lib.rs
@@ -2,19 +2,20 @@
 #![feature(llvm_asm)]
 
 pub mod hsmp;
-
 use hsmp::HSMP;
 
-pub struct SOC {}
+pub struct SOC {
+    hsmp: HSMP,
+}
 
 impl SOC {
     pub fn new() -> SOC {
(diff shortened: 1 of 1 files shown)


daym (approved):

src/soc/amd/rome/src/lib.rs
```diff
@@ -2,44 +2,46 @@
...
+    hsmp: HSMP,
+    topology: df::FabricTopology,
+}
 
 impl SOC {
     pub fn new() -> SOC {
-        Self {}
+        let hsmp = HSMP::new();
+        let topology = df::FabricTopology::new();
+        Self { hsmp, topology }
```
daym, line 16: Maybe it's not good to make it copy those

daym (changes requested):

src/soc/amd/rome/src/lib.rs
rminnich: so, looks like it was not copying it: I don't think FabricTopology has the Copy trait. I *think* the compiler did what is should. I changed it anyway.

daym (approved):

PR #406, opened 2021-04-08
Rome: Provide different stacks for different cores.

daym:
This makes `start.S` set up different stacks for different cores and threads.

It has been successfully tested.

The actual mechanism to enable physical cores in the first place is being documented by Paul, but he is not done yet (i.e. that part is not in this pull request). Therefore, this cannot be tested right now by people without AGESA access.

Note that it is my intention to have oreboot still only run on the bootstrap core. The other cores are only enabled so (1) Linux can use them and (2) we can set up MSRs for them.

The idea is that `_start` would check the `BSP` flag and then would `hlt` if it's not the bootstrap core--only for it to be later woken up for a short while by an interrupt (NMI, IPI or similar).

Note: TODO (later): adapt call `setup_acpi_tables(w, ?, <number of cores>)`

### DIFF
--- a/src/mainboard/amd/romecrb/src/bootblock.S
+++ b/src/mainboard/amd/romecrb/src/bootblock.S
@@ -79,9 +79,7 @@ gdt_end:
 3:
 	movb $0x42, %al
 	outb %al, $0x80
-	movw $0x3F8, %dx
   movb $0xAA, %al
-  call p4
 	// TOS contains a pointer to the gdt descriptor
 	popl %eax
 	// The low 4 bits of %eax are 3.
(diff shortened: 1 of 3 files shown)


daym:
With the newest commit, I get, after enabling the cores:

```
E1 AA
*E1 AA
*E0 AC
*E0 AC
*E0 AC
*E0 AC
*E0 AC
*E0 AC
*E0 AC
*00 AC
*00 AC
*00 AC
*00 AC
*00 AC
*00 AC
*00 AC
*00 AA
*00 AA
*00 AA
 00 AA
 00 18
 00 42
 00 18
 00 AA
 00 42
 00 BF
 00 18
 00 BE
 00 80
 00 76
 00 18
 00 BE
 00 A0
 00 BF
 00 BE
 00 F0
 00 BE
 00 30
 00 FF
 00 34
 00 FF
```

then it hangs (everything).

rminnich (approved):

PR #417, opened 2021-04-16
Clean up UART handling on Rome

daym:
This patch series cleans up the naming of the AMD UARTs in oreboot (to make it less confusing) and then makes romecrb use both UARTs again.

### DIFF
--- a/src/mainboard/asrock/a300m-stx/src/main.rs
+++ b/src/mainboard/asrock/a300m-stx/src/main.rs
@@ -14,7 +14,7 @@ use model::Driver;
 use print;
 use raw_cpuid::CpuId;
 use soc::soc_init;
-// use uart::amdmmio::AMDMMIO;
+// use uart::amdmmio::UART;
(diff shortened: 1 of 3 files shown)


src/mainboard/asrock/a300m-stx/src/main.rs
```diff
@@ -14,7 +14,7 @@ use model::Driver;
 use print;
 use raw_cpuid::CpuId;
 use soc::soc_init;
-// use uart::amdmmio::AMDMMIO;
+// use uart::amdmmio::UART;
```
rminnich, line 17: should we just delete this line?

rminnich (approved):

src/mainboard/asrock/a300m-stx/src/main.rs
daym: Probably. I thought about doing it, but I think Daniel is going to have to adapt A300m-stx to the new common AMD libs anyway...

PR #420, opened 2021-04-16
Factor out common parts from Rome mainboard to AMD common

daym:
This patchset moves common parts from mainboard/amd/romecrb directory to AMD common/boot directory, among others bootblock.S (but not start.S), consdebug, _asdebug and the payload runner.


src/soc/amd/common/boot/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+use model::Driver;
+use uart::i8250::I8250;
+mod acpi;
+use acpi::setup_acpi_tables;
+extern crate heapless; // v0.4.x
+use heapless::consts::*;
+use heapless::Vec;
+
+use core::ptr;
+// Until we are done hacking on this, use our private copy.
```
rminnich, line 18: Thiis comment probably no longer applies

src/soc/amd/common/boot/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+// Plan to copy it back later.
+global_asm!(include_str!("bootblock.S"));
+
+fn poke32(a: u32, v: u32) -> () {
```
rminnich, line 22: can we move these to arch?

src/soc/amd/common/boot/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+    unsafe { ptr::read_volatile(y) }
+}
+
+fn peekb(a: u64) -> u8 {
+    let y = a as *const u8;
+    unsafe { ptr::read_volatile(y) }
+}
+
+// Returns a slice of u32 for each sequence of hex chars in a.
+fn hex(a: &[u8], vals: &mut Vec<u64, U8>) -> () {
```
rminnich, line 70: This probably belongs in a library someday

src/soc/amd/common/boot/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+            b'h' => {
+                memb(w, line);
+            }
+            _ => {}
+        }
+    }
+}
+//global_asm!(include_str!("init.S"));
+
+pub fn boot(w: &mut impl core::fmt::Write, fdt_address: usize) {
```
rminnich, line 191: The more we separate this stuff ... the more I feel like passing the writer in as a paramater might be better? I don't know however.

rminnich (commented):
just a few thoughts.

src/soc/amd/common/boot/src/lib.rs
daym: Yes. I have to agree: Passing the writer explicitly every time isn't that bad.

Let's do that for now. We could always change over later anyway.

src/soc/amd/common/boot/src/lib.rs
daym: Probably--although maybe we can just use/pass the Writer anyway. Not that sureful to have

Anyway, for now, this patchset only moves stuff and doesn't change it if possible. That way, the move can be identified in the commit log later.

src/soc/amd/common/boot/src/lib.rs
daym: I don't like the API much--we definitely shouldn't be using it in general. I only kept them because of the one call in boot()--and even that I want to remove in favor of VolatileCell later.

See also https://www.ralfj.de/blog/2020/12/14/provenance.html for why.

src/soc/amd/common/boot/src/lib.rs
daym: Agreed--will/did remove in phase 2, in a new pull requested.

rminnich (approved):

PR #421, opened 2021-04-21
Factor out common parts from AMD mainboard "romecrb"

daym:
This patchset factors out more common parts from AMD mainboard "romecrb".

First, make I8250 and DebugPort actually own their respective IOPorts.

Then, add WriteToDyn to print as an extra fmt::Write implementation that is implemented on a `dyn Driver` rather than a on concrete implementation of `Driver`.

Then, make the `romecrb` mainboard return a list of possible text output drivers.
Use that in `main.rs` in order to find out which text output drivers to use.

Last, clean out obsolete comment about custom bootblock.S from src/soc/amd/common/boot/src/lib.rs .

### DIFF
--- a/src/mainboard/amd/romecrb/src/main.rs
+++ b/src/mainboard/amd/romecrb/src/main.rs
@@ -15,10 +15,9 @@ use print;
 use raw_cpuid::CpuId;
 use smn::{smn_read, smn_write};
 use soc::SOC;
-use uart::amdmmio::AMDMMIO;
-use uart::i8250::I8250;
 mod mainboard;
 use mainboard::MainBoard;
+use uart::i8250::I8250;
 mod fabric;
 use fabric::fabric;
 mod msr;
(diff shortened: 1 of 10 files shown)


rminnich (approved):

src/mainboard/amd/romecrb/src/main.rs
```diff
@@ -475,8 +451,7 @@ pub extern "C" fn _start(fdt_address: usize) -> ! {
 #[panic_handler]
 fn panic(info: &PanicInfo) -> ! {
     // Assume that uart0.init() has already been called before the panic.
-    let io = &mut IOPort;
-    let uart0 = &mut I8250::new(0x3f8, 0, io);
+    let uart0 = &mut I8250::new(0x3f8, 0, IOPort {});
```
rminnich, line 454: this kind of points out another problem: the IoPort{} probably ought to have a base, range, and stride.

e.g.
IoPort.New(0x3f8, 8, 1) -> 3f8 for 8 bytes with a stride of 1 (meaning use inb/outb)

for cf8/cfc:
ioport.New(0xcf8, 8, 4);

our ioport abstraction is lacking.

rminnich (approved):

PR #423, opened 2021-04-23
Rome: Init all the text output drivers automatically.

daym:
After the big refactoring, I forgot to call the Driver::init for all the text output drivers. That made the serial ports not print any output on the BMC serial on real hardware.

This patch adds it again.


orangecms (approved):
Looks good and works on the A300. PR for alignment coming up :)

PR #424, opened 2021-04-24
Rome: Remove setting of reserved bit in HPETCONFIG.

daym:
For some reason, we are setting a reserved bit in HPETCONFIG. If there's no good reason, we shouldn't.


orangecms:
Yup. The formatting things is nuts... it wants an empty line here

rminnich (dismissed):

rminnich (approved):

PR #425, opened 2021-04-24
align A300 with romecrb and add/fix some docs

orangecms:


orangecms:
Here is one issue: With the architectural switch to getting the drivers from `mainboard.rs`, the output is now significantly slower. We should investigate this.

daym:
I think we talked about this and the newest status is that it's fine to actually do for A300. Do you want to make this a real non-draft pull request instead?

daym (dismissed):

orangecms:
> I think we talked about this and the newest status is that it's fine to actually do for A300. Do you want to make this a real non-draft pull request instead?

Right, let me push my fixup and remove the draft status :)

PR #428, opened 2021-05-07
AMD Bootblock: Fix up.

daym:
`bootblock.S` had trashed %esp that `start.S` meticiously set up.

This patch rectifies that.

Also, clean up some comments and also make the rounding-up more resilient.


rminnich (approved):
good to see horrible esp go away :-)

PR #430, opened 2021-05-07
AMD Rome: Remove duplicate LDN setup.

daym:
This setup is already done in src/amd/common/boot, so don't do it again.

### DIFF
--- a/src/mainboard/amd/romecrb/src/main.rs
+++ b/src/mainboard/amd/romecrb/src/main.rs
@@ -415,14 +415,9 @@ fn start_bootstrap_core(fdt_address: usize) -> ! {
         msrs(w);
     }
     c00(w);
-    write!(w, "LDN is {:x}\r\n", peek32(0xfee000d0)).unwrap();
-    poke32(0xfee000d0, 0x1000000);
-    write!(w, "LDN is {:x}\r\n", peek32(0xfee000d0)).unwrap();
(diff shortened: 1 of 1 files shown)


src/mainboard/amd/romecrb/src/main.rs
```diff
@@ -415,15 +409,12 @@ fn start_bootstrap_core(fdt_address: usize) -> ! {
...
-    write!(w, "LDN is {:x}\r\n", peek32(0xfee000d0)).unwrap();
-    poke32(0xfee000d0, 0x1000000);
-    write!(w, "LDN is {:x}\r\n", peek32(0xfee000d0)).unwrap();
-    write!(w, "loading payload with fdt_address {}\r\n", fdt_address).unwrap();
 
     boot(w, fdt_address);
 
-    write!(w, "Unexpected return from payload\r\n").unwrap();
-    arch::halt()
+    loop {
```
orangecms, line 415: Could you elaborate a bit more on the `loop`? I'm not sure I understand why that would be necessary. Doesn't `halt()` guarantee that the machine stops?

daym: Oops, you are right, that‘s oreboot‘s halt fn which already uses the HLT instruction in an endless loop. I‘ll remove this chunk here on monday.

orangecms (approved):

PR #432, opened 2021-05-14
AMD: Add init_pics.

daym:
I don't know who thought relying on a traditional 1981 PIC in 2021 was a good idea--but here we go, we now enable it and then set it to ignore all external interrupts.


rminnich (approved):

rminnich:
DOS 1.0!!!
Have to have it!!!

rminnich (dismissed):

rminnich (approved):

PR #458, opened 2021-07-13
Add minimal preprocessor for assembly files

jhand2:
In order to support macro definitions and code reuse in oreboot
assembly, add a minimal preprocessor library. This library only supports
the "#include" and "#define" preprocessor directives and has no
toolchain-specific dependencies.


tools/rpp/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+    return out;
+}
+
+fn process_line(line: &str, ctx: &mut Context) -> Result<String, ParsingError> {
+    // Trim whitespace
+    let trimmed = line.trim();
+
+    if trimmed.starts_with("#") {
+        // For lines with keywords, remove comments
+        let replaced = ctx.eol_comment_re.replace_all(trimmed, "");
```
rjoleary, line 135: Strip comments before checking if the line starts with a '#'.

jhand2: Fixed this (just stripped all block comments before doing any parsing), although now the line numbers will be a little off for error reporting. Oh well, I added a todo, it's a problem for another day.

rjoleary:
Don't forget to run `cargo fmt`!

rjoleary:
This looks good. Just needs a `cargo fmt`.

jhand2:
> This looks good. Just needs a `cargo fmt`.

Thanks, done!

tools/rpp/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+    let mut parts = value.split_whitespace();
+    let key = match parts.next() {
+        Some(k) => k,
+        None => {
+            return Err(ParsingError::new(
+                String::from("#define has no name"),
+                ctx.line_num,
+            ))
+        }
+    };
```
LingMan, line 77: Looks like this could be written a bit shorter like this:
```rust
    let key = parts.next().ok_or_else(|| ParsingError::new(...))?;
```
Similarly in other places in this PR.

jhand2: Thanks for the suggestion! Done.

rjoleary (approved):

PR #465, opened 2021-08-06
Preprocess x86 bootblock asm files

jhand2:
Add the preprocess_asm! procedural macro for running asm files though
the preprocessor at compile-time.

Additionally, run the x86 bootblock asm files through the preprocessor so
they can share macro definitions.

### DIFF
--- a/tools/rpp_procedural/src/lib.rs
+++ b/tools/rpp_procedural/src/lib.rs
@@ -1,0 +1,52 @@
+use proc_macro::TokenStream;
+use std::env;
+use std::path::Path;
+
+#[proc_macro]
+pub fn preprocess_asm(input: TokenStream) -> TokenStream {
+    let mut rpp_ctx = rpp::Context::new();
+    let in_file = input.to_string();
+    let p = &in_file[1..in_file.len() - 1];
+
+    let curr_path = env::current_dir().unwrap();
+    let path = Path::new(&p);
+
+    // Create a path to the asm directory
+    let mut asm_dir = env::current_dir().unwrap();
+    asm_dir.push(path.to_path_buf());
+
+    // Change directory to the directory the asm file is in so all #include
+    // directives are relative to the root asm file.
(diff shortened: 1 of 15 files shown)


rjoleary (dismissed):

jhand2:
@rjoleary Good new! `println!("cargo:rerun-if-changed=...")` Just Works (TM) in procedural macros.

tools/rpp_procedural/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+        ),
+    };
+
+    let path_buf = path.to_path_buf();
+    let filename = &path_buf.file_name().unwrap().to_str().unwrap();
+
+    let mut out: String = rpp::process_file(&filename, &mut rpp_ctx).unwrap();
+
+    for f in rpp_ctx.processed_files {
+        println!("cargo:rerun-if-changed={}", f);
```
rjoleary, line 35: It looks like this is relative to the .S file and not to the crate.

rjoleary (dismissed):

tools/rpp_procedural/src/lib.rs
jhand2: Huh, now that you say this, I wouldn't expect this to actually work (see cargo implementation https://github.com/rust-lang/cargo/blob/bcd7fba64f8d3dcad5229ec75faeab018ec09c42/src/cargo/core/compiler/output_depinfo.rs#L80).

But it does work. I'll dig around to see why and if I can come up with something more robust.

jhand2: Alright well I think it just works because of an implementation quirk I don't understand. To play it safe, I used absolute filepaths.

rjoleary (approved):

jhand2:
Thanks @rjoleary! Feel free to merge this (or give me permissions so I can merge it 😄 )

PR #466, opened 2021-08-07
add Allwinner/sunxi Nezha

orangecms:
just a draft for visibility; nothing working thus far, just bare bones :skull_and_crossbones: 

note: the directory should be named `sunxi`, not `rvboards`


orangecms:
more references: https://github.com/smaeul/sun20i_d1_spl

orangecms:
https://github.com/michaelengel/xv6-d1 another port

xfel can init DRAM, so the first initial bit with just UART and a little print should be runnable right away

orangecms:
It prints _Welcome to oreboot_ \o/

src/drivers/uart/src/sunxi.rs
```diff
@@ -0,0 +1,170 @@
...
+
+const RETRY_COUNT: u32 = 100_000;
+
+/*
+ * see also https://github.com/smaeul/sun20i_d1_spl for reference
+ * files: nboot/main/boot0_head.c and drivers/serial.c
+ */
+
+// D1 user manual
+// https://dl.linux-sunxi.org/D1/D1_User_Manual_V0.1_Draft_Version.pdf
```
rminnich, line 18: if it is allowed, consider adding this to the oreboot datasheets repo.

src/drivers/uart/src/sunxi.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /// Returns a pointer to the register block
+    fn ptr(&self) -> *const RegisterBlock {
+        self.base as *const _
+    }
+}
+
+impl Driver for Sunxi {
+    fn init(&mut self) -> Result<()> {
```
rminnich, line 108: this might also someday move to the ctl function with the 'on' operation.

src/mainboard/sunxi/nezha/Makefile
```diff
@@ -0,0 +1,28 @@
+OREBOOT=$(abspath $(CURDIR)/../../../../)
+# RV64GCV, ackshully
+TARGET     = riscv64gc-unknown-none-elf
+QEMU       ?= qemu-system-riscv64
+QEMU_FLAGS += -m 1g -machine sifive_u,start-in-flash=true -nographic -device loader,addr=0x20000000,file=${IMAGE} -bios none -smp 4
+XFEL       ?= xfel
+IMAGE      ?= target/$(TARGET)/release/image.bin
+# DRAM space starts at 0x40000000, see manual p34
+MEMADDR    ?= 0x40000000
```
rminnich, line 9: Do you really want the ?= operator here? is it really possible to change MEMADDR?

src/mainboard/sunxi/nezha/Makefile
```diff
@@ -0,0 +1,28 @@
...
+include ../../../../Makefile.inc
```
rminnich, line 10: Note that we could make this, for example, $(OREBOOT)/Makefile.inc

src/mainboard/sunxi/nezha/fixed-dtfs.dts
```diff
@@ -0,0 +1,46 @@
...
+                description = "Boot Blob and Romstage";
+                size = <0x80000>; // 512KiB
+                file = "$(TARGET_DIR)/bootblob.bin";
+            };
+            area@1 {
+                description = "Fixed DTFS";
+                size = <0x80000>; // 512KiB
+                file = "$(TARGET_DIR)/fixed-dtfs.dtb";
+            };
+            area@2 {
```
rminnich, line 23: TODO: the tool should be able to size areas dynamically. The fixed sizes are because we don't yet walk to fdt nodes ... not your problem, just mentioning it.

src/mainboard/sunxi/nezha/link.ld
```diff
@@ -0,0 +1,64 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+ENTRY(_boot);
+
+SECTIONS
+{
+    . = 0x40000000;
```
rminnich, line 29: That this is really fixed, so the ?= above for MEMADDR probably ought to be = instead.

src/mainboard/sunxi/nezha/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+use soc::ccu::CCU;
+use soc::gpio::GPIO;
+use uart::sunxi::Sunxi;
+use uart::sunxi::UART0;
+
+global_asm!(include_str!("../start.S"));
+
+// hart = hardware thread (something like core)
+#[no_mangle]
+pub extern "C" fn _start_here() -> ! {
```
rminnich, line 19: could we just call this _start to be like the other boards?

src/mainboard/sunxi/nezha/start.S
```diff
@@ -0,0 +1,46 @@
+#define REGBYTES        4
+/*mxstatus*/
+#define    EN_THEADISAEE    (0x1 << 22)
+
+.section ".bootblock.boot", "ax", %progbits
```
rminnich, line 5: fantastic! minimal assembly!

src/soc/sunxi/d1/Makefile
```diff
@@ -0,0 +1,4 @@
+OREBOOT=$(abspath $(CURDIR)/../../../../)
+include ../../../../Makefile.inc
```
rminnich, line 2: this could also be $(OREBOOT)/Makefile.inc

src/soc/sunxi/d1/src/ccu.rs
```diff
@@ -0,0 +1,91 @@
...
+    fn pwrite(&mut self, data: &[u8], _offset: usize) -> Result<usize> {
+        /*
+        'outer: for (sent_count, &c) in data.iter().enumerate() {
+            return Ok(sent_count);
+        }
+        */
+        Ok(data.len())
+    }
+
+    fn ctl(&mut self, __d: DeviceCtl) -> Result<usize> {
```
rminnich, line 76: I think these can start being Ok() instead of NOT_IMPLEMENTED?

rminnich (approved):
If this is printing a startup message, you can ignore my nits, commit it, and loop back around later and fix it. Some things, like the use of ../../../../ instead of $(OREBOOT), can be changed globally in another PR. Other things, like changing init to be ctl(On), can be done another time. I think this is fine.

src/drivers/uart/src/sunxi.rs
orangecms: It has a watermark, says something about confidential etc, though anyone could also sign up for the Allwinner dev portal and download it. I'd rather ask Allwinner, though on the other hand, the linux-sunxi wiki has a bunch of resources on it that I'd not want to spoil. I have very mixed feelings.

src/mainboard/sunxi/nezha/Makefile
orangecms: Yea I thought about it. It's temporary anyway until we can boot from flash (and actually flash). One idea was to extend the Makefile with a little `make load` where we could load something else somewhere for testing bits and pieces gradually. I'll think about it.

PR #468, opened 2021-08-14
Add features to the rust preprocessor library

jhand2:
To support more full-featured preprocessing, add the following features
to the rpp and rpp_procedural crates:
* Fix block comment bug that led to wrong line numbers in parsing errors
* Add support to preprocess_asm! to inject macro definitions

Additionally, add some basic tests for the rpp and rpp_procedural crates.

### DIFF
--- a/tools/rpp/Cargo.lock
+++ b/tools/rpp/Cargo.lock
@@ -10,6 +10,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "cfg-if"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
+
+[[package]]
+name = "getrandom"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7fcd999463524c52659517fe2cea98493cfe485d10565e7b0fb07dbba7ad2753"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "wasi",
+]
+
+[[package]]
+name = "libc"
+version = "0.2.99"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a7f823d141fe0a24df1e23b4af4e3c7ba9e5966ec514ea068c93024aa7deb765"
+
+[[package]]
 name = "memchr"
 version = "2.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


rminnich:
you're following the path to hell with the ifdef, are you really sure?

jhand2:
> you're following the path to hell with the ifdef, are you really sure?

Here's the specific scenario I'm trying to solve:
Code to drop to protected mode or enter long mode requires knowing the location of code segments, data segments, and page table. These segments are at different locations depending on which bootblock you use (ROM-only or RAM). So I end up with essentially the following:

```
#ifdef X86_ROM_GDT
#define DAT32 0x20
#define CODE32 0x18
#define CODE64 0x28
#define PAGE_TABLE_ADDR pml4 // Page table is defined in the bootblock
#else 
#define DAT32 0x10
#define CODE32 0x08
#define CODE64 0x18
#define PAGE_TABLE_ADDR 0x7e000
#endif
```

There is one alternative I can think of: Define these things in rust instead and pass them into the `preprocess_asm!` macro. This probably mostly works, but it's a little precarious because you can only pass string literals as inputs to `preprocess_asm!`. But it's still an option.

jhand2:
> > you're following the path to hell with the ifdef, are you really sure?
> 
> Here's the specific scenario I'm trying to solve:
> Code to drop to protected mode or enter long mode requires knowing the location of code segments, data segments, and page table. These segments are at different locations depending on which bootblock you use (ROM-only or RAM). So I end up with essentially the following:
> 
> ```
> #ifdef X86_ROM_GDT
> #define DAT32 0x20
> #define CODE32 0x18
> #define CODE64 0x28
> #define PAGE_TABLE_ADDR pml4 // Page table is defined in the bootblock
> #else 
> #define DAT32 0x10
> #define CODE32 0x08
> #define CODE64 0x18
> #define PAGE_TABLE_ADDR 0x7e000
> #endif
> ```
> 
> There is one alternative I can think of: Define these things in rust instead and pass them into the `preprocess_asm!` macro. This probably mostly works, but it's a little precarious because you can only pass string literals as inputs to `preprocess_asm!`. But it's still an option.

The alternative would look something like this:


in `src/arch/x86/x86_64/src/lib.rs`
```
macro_rules! ROM_DAT32 { () => { 0x20 } };
macro_rules! ROM_CODE32 { () => { 0x18 } };
macro_rules! ROM_CODE64 { () => { 0x28 } };
macro_rules! ROM_PAGE_TABLE_ADDR { () => { "pml4" } };

macro_rules! RAM_DAT32 { () => { 0x20 } };
macro_rules! RAM_CODE32 { () => { 0x18 } };
macro_rules! RAM_CODE64 { () => { 0x28 } };
macro_rules! RAM_PAGE_TABLE_ADDR { () => { "0x7e000" } };
```

in a mainboard file like `src/mainboard/emulation/qemu-fsp/src/main.rs`
```
global_asm!(preprocess_asm!(
    "../../../arch/x86/x86_64/src/bootblock_nomem.S",
    concat!("DAT32=", ROM_DAT32!()),
    concat!("CODE32=", ROM_CODE32!()),
    concat!("CODE64=", ROM_CODE64!()),
    concat!("PAGE_TABLE_ADDR=", ROM_PAGE_TABLE_ADDR!())
));
```

jhand2:
Alright, ifdef is dead, long live ifdef.

rminnich (approved):

rminnich (approved):

PR #470, opened 2021-08-26
Drop to protected mode to call 32-bit FSP

jhand2:
Most FSP blobs are 32-bit binaries. Add functionality to drop into
protected mode to call FSP init functions.

In order to support this, add the following:
* Refactor fspsdk to pull out common bindings to the uefi and fsp-common
crates.
* Refactor x86 bootblock so assembly to enter long mode can be reused
in the FSP-calling code.


rminnich:
This seems ready to go in?

rminnich (approved):

PR #472, opened 2021-09-10
Run all the makefiles for cibuild

rminnich:
Rather than have the top level list of which boards build and which do not, put
a cibuild target in each makefile. It is then easy to turn it on and off in the makefile;
further, the cibuild targets can be individually tested before pushing boards.

If this model works, then we can have simple toplevel targets for mainboards,
and put the logic for a given board in that board's makefile.

This will avoid all the confusion around exceptions that we have today, as well as
supporting a general principle: all make targets should work at top level
and in each mainboard directory.


jhand2 (dismissed):
> all make targets should work at top level and in each mainboard directory.

+1, I like this principle

orangecms (approved):
Yup, good idea!

jhand2:
There's some changes to the fspsdk submodule that makes the qemu-fsp board sad. I think the following should fix the problem:

```
cd 3rdparty/fspsdk
git fetch origin qemu_fsp_x64
git checkout 037b610dfa90eb4f5d38552e10b1f98a58427bc2
cd ../..
<commit the submodule changes>
```

jhand2 (approved):

LeanSheng (approved):

azure-pipelines.yml
```diff
@@ -70,8 +70,8 @@ jobs:
   - script: |
       (
         cd src/mainboard/emulation/qemu-fsp
-        timeout 120s make run | tee serial
-        grep 'Running payload' serial
+        echo 'NO: timeout 120s make run | tee serial'
+        echo 'NO: grep Running.payload serial'
```
jhand2, line 74: The fsp job is just failing because of the fspsdk submodule change

rminnich: well what we could do ... just take this as is for now and start turning stuff back on.
I'd rather do that, as it might unblock a bunch of other PRs as well.

LeanSheng: Sounds good to me.

LeanSheng (approved):

azure-pipelines.yml
jhand2: Works for me. I'll re-enable this in a little bit

PR #474, opened 2021-09-17
upxtreme: Initial commit

LeanSheng:
Signed-off-by: Lean Sheng Tan <tanleansheng@outlook.com>

### DIFF
--- a/src/mainboard/aaeon/upsquared/fixed-dtfs.dts
+++ b/src/mainboard/aaeon/upsquared/fixed-dtfs.dts
@@ -6,7 +6,7 @@
 
     flash-info {
         compatible = "ore-flashinfo";
-        board-name = "UP Sqared";
+        board-name = "UP Squared";
         category = "aaeon";
         board-url = "https://up-board.org/upsquared/";
         areas {
--- a/src/mainboard/aaeon/upxtreme/Cargo.lock
+++ b/src/mainboard/aaeon/upxtreme/Cargo.lock
@@ -19,6 +19,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "arch"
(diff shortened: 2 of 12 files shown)


src/mainboard/aaeon/upsquared/fixed-dtfs.dts
```diff
@@ -6,9 +6,9 @@
 
     flash-info {
         compatible = "ore-flashinfo";
-        board-name = "UP Xtreme";
+        board-name = "UP Sqared";
```
rjoleary, line 9:
```suggestion
        board-name = "UP Squared";
```

src/mainboard/aaeon/upsquared/fixed-dtfs.dts
```diff
@@ -6,9 +6,9 @@
 
     flash-info {
```
rjoleary, line 7: I think you should delete this file -- it's upsquared while everything else is upxtreme.

rjoleary:
The CI system is failing with this format error:

```
Diff in /home/vsts/work/1/s/src/mainboard/aaeon/upxtreme/src/main.rs at line 14:
 
 use rpp_procedural::preprocess_asm;
 
-use fsp_common as fsp;
 use fsp_cfl_sys as fsp64;
+use fsp_common as fsp;
```

src/mainboard/aaeon/upsquared/fixed-dtfs.dts
LeanSheng: i fix this so it wont have duplicate name with the upxtreme board. Is this fine with you?

rminnich (approved):

PR #475, opened 2021-09-24
Continue to build on the new makefile layout

rminnich:
In the early days, we had not Makefiles; our build file was xcargo
or similiar.

Our first pass at Makefiles involved a Makefile at the top with
Makefiles only for mainboards. The Makefile at the top had targets
not available in most places, even in mainboards, and that has proven
very inconvenient.

This commit instantiates a new set of rules for how to use make in
oreboot. The goal is to avoid the problems we saw in previous versions
of both oreboot and coreboot.

The rules are simple:
1. any directory with a Cargo.toml should have a Makefile
2. every Makefile should have a common set of rules for
   format
   clippy
   build
   default
3. When possible, these rules should be defined in a
   top level Makefile.inc
4. mainboards have an extended Makefile designed to
   supporting building a target
5. At two level, there are two lists:
   ALLMAKEFILES
   MAINBOARDMAKEFILES
6. For all directories containing a Makefile,
   it shall be possible to run make with the same
   target as at top level
   e.g, make format at top level should work;
   make format should work for any directory with
   a Makefile

This is not quite complete. But a few more steps should get us there.


rminnich:
note this requires setting OREBOOT in the environment, and I don't recall how to do that in azure.

Makefile.inc
```diff
@@ -1,5 +1,6 @@
-# This Makefile is included by src/mainboards/VENDOR/BOARD/Makefile
-TOP         = $(abspath $(CURDIR)/../../../..)
+# This file is included by all Makefiles. The user is required to set
+# the path to OREBOOT.
+TOP         = $(OREBOOT)
```
jhand2, line 3: Add something to the top level readme under "Setting up the toolchain" recommending adding export OREBOOT=... to bashrc.

Makefile
```diff
@@ -63,6 +63,20 @@ CRATES := \
 	$(wildcard */*/*/*/Cargo.toml) \
 	$(wildcard */*/*/*/*/Cargo.toml)
 
+ALLMAKEFILE := \
```
rjoleary, line 66: Won't this go into the 3rdparty directory too which might get weird?

Makefile
```diff
@@ -63,6 +63,20 @@ CRATES := \
...
+	$(wildcard */Makefile) \
+	$(wildcard */*/Makefile) \
+	$(wildcard */*/*/Makefile) \
+	$(wildcard */*/*/*/Makefile) \
+	$(wildcard */*/*/*/*/Makefile)
```
rjoleary, line 71: `$(shell find -name Makefile -type f)

Makefile.inc
rjoleary: Is it possible to automatically determine TOP without the user setting any variables?

Makefile
rminnich: we actually need to make that top level [a-z]* it turns out. The find will find too many things. but it will ignore 3rdparty

Makefile
rminnich: yes that's bad.

Makefile.inc
rminnich: done. It's automatic now.

orangecms:
Turning off Azure can simply be done in the GitHub settings, should we want to disable it completely for the time being. :)

Makefile.inc
```diff
@@ -36,6 +37,21 @@ $(shell echo "rustc --version  = $(shell rustc --version)" 1>&2)
 $(shell echo "rustup --version = $(shell rustup --version)" 1>&2)
 $(shell echo "**********" 1>&2)
 
+# The default target: build the board.
+mainboard: $(IMAGE)
```
jhand2, line 41: Hmmm. We're include this Makefile.inc file for every crate, but most of these variables only apply to (and are set for) mainboards. So these targets will do weird stuff on all the non-mainboard makefiles.

payloads/Makefile
```diff
@@ -0,0 +1,2 @@
+OREBOOT=../
```
jhand2, line 1: All of these declarations need to be of the form `$(abspath $(CURDIR)/..)`. Just using plain relative paths won't work.

Makefile.inc
rminnich: we can fix that, I think, on the next pass. Good point.

payloads/Makefile
rminnich: fixed

.github/workflows/fsp-qemu.yml
```diff
@@ -75,7 +75,7 @@ jobs:
         (
           cd qemu/build-riscv64
           ../configure --target-list=riscv64-softmmu
-          make -j16
+          make -j16 mainboard
```
orangecms, line 78: this is the skipped make for QEMU itself, which wouldn't have a `mainboard` target

.github/workflows/fsp-qemu.yml
```diff
@@ -113,7 +113,7 @@ jobs:
...
```
orangecms, line 116: same as above; QEMU itself just needs `make`

rminnich: fixed.

.github/workflows/fsp-qemu.yml
rminnich: fixed.

jhand2 (approved):
It's so beautiful :')

PR #481, opened 2021-09-27
Cleanupbeagle

rminnich:


src/soc/starfive/jh7100/src/iopad.rs
```diff
@@ -149,7 +149,7 @@ pub const syscon_iopad_ctrl_register102_REG_ADDR: u32 = SYSCON_IOPAD_CTRL_BASE_A
 pub const syscon_iopad_ctrl_register103_REG_ADDR: u32 = SYSCON_IOPAD_CTRL_BASE_ADDR + 0x19C;
 pub const syscon_iopad_ctrl_register104_REG_ADDR: u32 = SYSCON_IOPAD_CTRL_BASE_ADDR + 0x1A0;
 
-pub fn _SET_SYSCON_REG_register0_SCFG_gpio_pad_ctrl_0(v: u32) -> () {
+pub fn _SET_SYSCON_REG_register0_SCFG_gpio_pad_ctrl_0(v: u32)  {
```
fotonick, line 152: I'm surprised that clippy is happy with two spaces between the close paren and the open curly brace. It seems better to have one space here.

azure-pipelines.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       make --keep-going test
     displayName: 'Run rust tests'
   - script: |
-      echo let us wait until cargo fix or rustfix actually works make --keep-going clippy
+      make --keep-going clippy
     displayName: 'SKIP clippy linter'
```
fotonick, line 32: Update display name since you're no longer skipping?

src/soc/starfive/jh7100/src/clock.rs
```diff
@@ -256,7 +256,7 @@ pub fn _ENABLE_CLOCK_clk_cpundbus_root_() {}
 pub fn _SWITCH_CLOCK_clk_cpundbus_root_SOURCE_clk_osc_sys_() {
     let mut v = peek32(clk_cpundbus_root_ctrl_REG_ADDR);
     v &= !(0x3 << 24);
-    v |= (0x0 & 0x3) << 24;
+    v |= 0 << 24;
```
fotonick, line 259: I know you're just satisfying clippy, so maybe it's out of scope, but in many of these places, there's further simplication: 0 << 24 == 0. v | 0 == v, so this line can be removed entirely. It also feels like we're missing an opportunity to use a Rust zero-cost abstraction here for safety. Perhaps there's a good bitflags crate to use here to reduce opaque magic numbers everywhere.

fotonick (changes requested):
Thanks for doing the good work of satisfying clippy. I do think the spaces and displayName comments should be addressed.

azure-pipelines.yml
orangecms: or just revert that one commit that did the change

src/soc/starfive/jh7100/src/clock.rs
rminnich: I agree and those changes are coming.

rminnich: This was the output of a script pipeline from C. The C itself was a crime against nature. So it's a bit of a mess, but also way too much code to do by hand, and I"m not smart enough to do it with anything but sed.

src/soc/starfive/jh7100/src/iopad.rs
rminnich: rustfmt was my friend.

orangecms (approved):

Issue #452, opened 2021-06-26
QEMU-FSP: Use the SplitFspBin.py tool instead of `git apply`


rjoleary:
Currently, inside the `firsttime_fsp` recipe of the main Makefile, we `git apply` a patch onto fspsdk to change the FSP_BASE address. This only work because we re-compile the QEMU-FSP from source.

Instead, we should use the SplitFspBin.py Python program to relocate the FSP binary to a new base after it is compiled. See: https://github.com/universalpayload/fspsdk/blob/qemu_fsp_x64/IntelFsp2Pkg/Tools/SplitFspBin.py

aquadara:
I believe PR #482 (to rebase) and PR #484 to (to split thereafter) addressed this - closing out.

PR #482, opened 2021-09-30
QEMU-FSP: Use the SplitFspBin.py tool per issue #452

aquadara:

There's a bit more to do for this PR, specifically:
- Decide if "We may want to parse the file to derive these, or at the very least verify they match up..." can be left TODO
- Understand why, while QEMU runs fine with this code, the last 4 bytes of each component don't line up per `hexdump` when comparing before (FLASH_BASE patch and build) and after (no patch, build and rebase with SplitFspBin.py)

RebaseDiff

I don't know what the last 4 bytes of each section is yet. Might not matter, might be a big deal, might be a limitation in SplitFspBin.py. Will take to the group tomorrow.


orangecms:
Those last bytes seem to be this to me:
https://github.com/tianocore/edk2/blob/master/IntelFsp2Pkg/Tools/SplitFspBin.py#L688

aquadara:
> Those last bytes seem to be this to me: https://github.com/tianocore/edk2/blob/master/IntelFsp2Pkg/Tools/SplitFspBin.py#L688

Seemed like a good guess, but I *think* I was able to prove otherwise. Print statements suggest the offset touched by that line is never the offset of the diff. Furthermore, I changed the bytes manually and reran the tool - the bytes went unchanged. I think ultimately the tool doesn't touch those bytes in its rebasing effort, but the FLASH_BASE patch has consequences in them that take effect at build time. I'm gonna give it a look with fresh eyes tomorrow but we can discuss more at the mtg if I don't figure it out.

rjoleary (approved):

orangecms (approved):

PR #483, opened 2021-10-01
Make the mainboard when running 'make run' for QEMU

aquadara:

### DIFF
--- a/Makefile.qemu.inc
+++ b/Makefile.qemu.inc
@@ -8,7 +8,9 @@ run: $(IMAGE)
 	$(QEMU) $(QEMU_FLAGS)
 
 trace: $(IMAGE)
+	make mainboard
 	$(QEMU) -d guest_errors,in_asm,int $(QEMU_FLAGS)
 
 gdb: $(IMAGE)
+	make mainboard
 	$(QEMU) -d guest_errors -s -S $(QEMU_FLAGS)
--- a/Makefile.qemu.inc
+++ b/Makefile.qemu.inc
@@ -3,11 +3,11 @@
 
 MONITOR ?= "/dev/null"
 
-run: mainboard $(IMAGE)
+run: mainboard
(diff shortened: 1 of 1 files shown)


Makefile.qemu.inc
```diff
@@ -4,6 +4,7 @@
 MONITOR ?= "/dev/null"
 
 run: $(IMAGE)
```
jhand2, line 6: Hmmm. I know this technically works, but seems odd. It only works because Makefile.qemu.inc and Makefile.mainboard.inc happen to both be included by every board in  `src/mainboard/emulation`.

Also the mainboard target is just `mainboard: $(IMAGE)` so the run target depending `$(IMAGE)` should already do the trick.

jhand2 (changes requested):

Makefile.qemu.inc
aquadara: So the is off the back of the Makefile changes that went in today, which appear to have broken `make run` triggering a rebuild of the mainboard if there were changes therein. On the call this was the suggested workaround, but one thing I can say for sure is that it wasn't working for me and Titus as-is.

jhand2: out of curiosity, does `run: $(IMAGE) mainboard` fix the problem, or no?

jhand2: Looks like that works for me. So as a small cleanup on this PR I'd prefer that to explicitly calling `make mainboard`

jhand2: Oh another think to try. In the emulation/*/Makefile files, switching the include order seems to also work. Including mainboard.inc before qemu.inc.

aquadara: Yep yeah agree - just bad at `make`. Fixed!

jhand2 (dismissed):

rminnich (approved):

PR #484, opened 2021-10-01
After rebasing the FSP binary, split it again

aquadara:


jhand2:
Can you add to the PR description why this is necessary? And also probably in a comment in the rs file.

In general, it's good to have PR descriptions/commit messages with enough detail for folks to understand why a change was make a long time ago.

aquadara:
> Can you add to the PR description why this is necessary? And also probably in a comment in the rs file.
> 
> In general, it's good to have PR descriptions/commit messages with enough detail for folks to understand why a change was make a long time ago.

So this is still a work in progress, but it was suggested on our call per https://github.com/slimbootloader/slimbootloader/blob/b24fd2759cdd3a484602a9a2221487e421fec838/BuildLoader.py, wherein all rebases are followed by splits. I don't believe it's necessary due to some hardcoding elsewhere in our codebase, but I think it is ultimately desired.

Still not quite sure how to test this either.

PR #485, opened 2021-10-03
Fix macos build

fotonick:
With the first patch to determine host target triple, builds seem to work on macOS. I was able to run qemu-aarch64, qemu-armv7, and qemu-riscv successfully. I wasn't able to get any of the x86 targets working, though I'm not sure if they're supposed to.

The second patch just helps everyone coming in with old rustup versions.


orangecms:
This doesn't quite seem to work yet...
```
error: error: The argument '--default-toolchain <default-toolchain>' requires a value but none was supplied
```

For macOS, we can introduce a CI check now on GitHhub Workflows:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

fotonick:
Sorry, fixed. Agreed that should look at checking macOS builds as part of the CI. Will have to find time.

I just noticed that building clippy is now the long pole of the CI, which is hilarious to me. I should try to find time to look into that.

rjoleary (approved):

PR #486, opened 2021-10-04
Add renode support to oreboot

philhug:
Use `make run_renode` to invoke renode
Enabled for sifive unleashed board


rminnich (approved):
This is *really* nice and you used the new makefile structure very effectively. Thanks very much!

PR #489, opened 2021-10-16
Add some Cargo.lock files to .gitignore

Rwantare:
I find it irritating to switch branches locally but having to
repeatedly revert or stash .lock changes I didn't mean to make.

It was discussed that the ones in src/mainboard should be kept.
Any others? src/lib? src/drivers?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,7 @@ oreboot.asm
 target
 **/*.rs.bk
 debug.log
+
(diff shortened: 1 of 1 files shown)


orangecms:
Wasn't it the other way around - have the obes for drivers, libs etc crates in the repo, and not the ones for mainboards? 🤔

Rwantare:
I think now that src/mainboards should be enough.

PR #501, opened 2021-11-05
FSP: Organize UPD parameters by arch instead of board

jhand2:
Refactor the boards which use FSP so that the FSP UPD structures are
kept in architecture-specific crates instead of board-specific crates.
This allow boards to share the same parameters for the same CPU
generation.

### DIFF
--- a/src/vendorcode/fsp/qemu/src/lib.rs
+++ b/src/vendorcode/fsp/qemu/src/lib.rs
@@ -2,6 +2,7 @@
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
 #![no_std]
+#![allow(clippy::zero_ptr)]
 
 // Rust types are used instead of generated ones.
 #[repr(C)]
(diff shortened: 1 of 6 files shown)


src/vendorcode/fsp/qemu/src/lib.rs
```diff
@@ -22,3 +23,59 @@ macro_rules! blob_macro {
...
+            UnusedUpdSpace0: [0; 49],
+            ReservedFspmUpd: [0; 4],
+        },
+        UnusedUpdSpace1: [0u8; 2],
+        UpdTerminator: 0x55AA, // ???
+    }
+}
+
+pub fn get_fsps_upd() -> FSPS_UPD {
+    // TODO: This struct has to be aligned to 4.
```
rminnich, line 61: Titus looked and this should be OK because that is C, and Rust should assume C alignment and ordering rules apply?

rminnich (commented):
we are unsure the TODO about alignment is needed.

rminnich:
btw, this is really a big improvement.

src/vendorcode/fsp/qemu/src/lib.rs
jhand2: Would be happy to hear this in the case 😄  I just retained this comment from the original implementation. But if we think it's irrelevant, happy to remove it

rjoleary: I think we don't need the TODO anymore. It's repr C and packed.

jhand2: Confirmed that this is the case: https://doc.rust-lang.org/reference/type-layout.html

"The alignment of the struct is the alignment of the most-aligned field in it."

So removed.

rminnich (approved):

PR #507, opened 2021-11-13
Fix ast build

edtanous:
Summary of changes:
Remove Cargo.locks that snuck in.
Rearrange ast25x0 into new pattern, separating mainboard folder from
SOC.
Add fixed-dtfs.dts (mostly copied from HiFive unleased)
Comment out now-unavailable uarts
Change arm rust toolchain from armv7r (realtime) to armv7a (application)
Change ast25x0 to use armv7a-none-eabihf toolchain.  Previously used
arm-non-eabihf toolchain no longer seems to exist.
Add nop method to arch arm
sed /s/registers/tock_registers/g and remove registers dependency
Remove unused SPICheck device, which doesn't seem to be used, and
doesn't compile in this version of rust.
Fix TimerEnable to use the appropriate struct member (not clear how this
could've ever compiled).
Comment out SiFive uart reference in main.rs.  Needs more looking into.

Tested:
cd src/mainboards/ast/ast25x0 && make mainboard now succeeds

V2 removes ciclippy;  AST code is far from ready for static analysis.


src/soc/aspeed/ast25x0/src/asmram.rs
```diff
@@ -128,7 +127,7 @@ fn peek(a: u32) -> u32 {
     let y = a as *const u32;
     unsafe { ptr::read_volatile(y) }
 }
-pub fn ram(w: &mut print::WriteTo) -> () {
+pub fn ram(w: &mut impl core::fmt::Write) -> () {
```
rminnich, line 130: I wonder if it is this way for space reasons? I no longer recall. But I also think that we can make print::WriteTo fit it might be better?

rminnich (dismissed):
fantastic. Let's make it part of the build test on the next PR. Needs a simple makefile change.

src/soc/aspeed/ast25x0/src/asmram.rs
edtanous: I have no context for the difference between these two, but there wasn't anywhere in soc where we used print::WriteTo that I could find;  I copied this core::fmt::Write call from soc/amd/common/boot/src/lib.rs;

https://github.com/oreboot/oreboot/blob/c66f625548f3439282edc95d89fa2a2f33799c21/src/soc/amd/common/boot/src/lib.rs#L80

Happy to change to whatever.

src/soc/aspeed/ast25x0/Makefile
```diff
@@ -0,0 +1,4 @@
+OREBOOT=$(abspath $(CURDIR)/../../../../)
+include ../../../../Makefile.inc
+ciclippy: clippy
```
edtanous, line 3: I'm assuming this is the line where I flew too close to the sun and turned on clippy ci for this long before it was ready for it.

edtanous:
> fantastic. Let's make it part of the build test on the next PR. Needs a simple makefile change.

ACK

edtanous:
I just spent an hour trying to make clippy happy, and it was a long and deep rabit hole.  Is there any way to just disable it for ast2500 for the moment?

Rwantare:
> I just spent an hour trying to make clippy happy, and it was a long and deep rabit hole. Is there any way to just disable it for ast2500 for the moment?

Something like `cargo clippy -Zbuild-std=core --target armv7a-none-eabihf --fix --allow-dirty` can automatically handle some changes for you. But it's unclear why you're dealing with clippy to begin with because the ast2500 Makefile has `ciclippy: skipclippy`

@rjoleary is this a bug?

edtanous:
> > I just spent an hour trying to make clippy happy, and it was a long and deep rabit hole. Is there any way to just disable it for ast2500 for the moment?
> 
> Something like `cargo clippy -Zbuild-std=core --target armv7a-none-eabihf --fix --allow-dirty` can automatically handle some changes for you

I tried that, it fixed some of the issues.  There's a lot more.

>. But it's unclear why you're dealing with clippy to begin with because the ast2500 Makefile has `ciclippy: skipclippy`
> 
> @rjoleary is this a bug?

"skipclippy" I think is the magic I was looking for.  Keep in mind, ast2500 is both a board and an soc at this point, and the soc folder was missing it.

src/mainboard/ast/ast25x0/Cargo.lock
```diff
@@ -1,323 +0,0 @@
-# This file is automatically @generated by Cargo.
```
Rwantare, line 1: I think Cargo.lock files in src/mainboard/** should be checked in

edtanous: I thought we talked about this last Friday and said the opposite, but clearly I misunderstood.  Will fix shortly.

rminnich (approved):

rminnich (approved):

rminnich (approved):

Issue #504, opened 2021-11-07
Replacing Assembly Preprocessor with Rust Built-in Features


retrage:
I'm new to this project, so correct me if I missed something.
As far as I know, PR #458 introduced rpp for preprocessing assembly files. The primary motivation is to improve assembly code reusability because Rust does not provide its macro system to assembly.

However, I cannot find any reason to employ the C preprocessor-like syntax for the project written in Rust. I think it is better to use rich Rust's compile-time features for preprocessing. The current primary usage of `preprocess_asm!()` is to define constant value macros and include other files. It can be replaced by nightly Rust's built-in feature because `global_asm!()` accepts named parameters. Here is my PoC:
https://github.com/retrage/oreboot/tree/asm-const-poc

The macro-defined values are defined in [src/consts/asm.rs](https://github.com/retrage/oreboot/blob/asm-const-poc/src/arch/x86/x86_64/src/consts/asm.rs). The macro references are replaced with Rust named parameters (e.g. `Pse` -> `{pse}`).
The corresponding const values are given as arguments of `global_asm!()`.
```rust
global_asm!(
    include_str!("../../../../arch/x86/x86_64/src/bootblock_nomem.S"),
    pse = const asm::PSE,
    // snip
    options(att_syntax)
);
```
This design also allows us to use other Rust const features like `const fn` without additional efforts.

The main drawback of this idea is that the number of `global_asm!()` arguments may be large, and the code looks redundant. We can hide the arguments by defining a macro that uses default values.

I can provide PR with more improvements if anyone is interested in my idea.

orangecms:
Oh wow, that's great, thank you a lot! It makes maintenance a lot easier with otherwise duplicated definitions. We currently have a bug with kernel execution in the q35 QEMU test port, which I suspect to be related to some divergence. With your ideas here, that would be totally mitigated. :)

rminnich:
Nice to have you here. Please submit your PR. We are still learning and welcome PRs like this!

jhand2:
I'm not overly attached to rpp, so ya if there's a more idiomatic rust way to do this I'm all for it 😊

jhand2:
Caveat that I'm a little weary of using nightly rust features because it would be nice to move to a stable tool chain. I don't know enough about this feature to know if/when it's headed to stable.

retrage:
Thanks, so I'll submit a PR.

retrage:
Close this issue because my PR #513 is merged.

PR #513, opened 2021-11-23
Replace rpp with global_asm Named Parameters

retrage:
As I proposed in issue #504, we can replace the assembly preprocessor `rpp` with the Rust built-in features. This PR is based on my PoC https://github.com/retrage/oreboot/tree/asm-const-poc.

The main change from the PoC branch is introducing a new macro. I added the `const_asm!()` macro to wrap the named parameter boilerplate code.
By the way, I think the name `const_asm` is too general and does not describe the role well. Could you give me comments on the name?


daym:
About `const_asm` name, I can see it's difficult to find a name for that. Maybe something like `asm_in_arch_env` ?

src/arch/x86/x86_64/src/bootblock.S
```diff
@@ -184,7 +189,7 @@ gdtptr16:
 
 	movl	%cr0, %eax
 	andl	$0x7FFAFFD1, %eax /* PG,AM,WP,NE,TS,EM,MP = 0 */
-	orl	$(CD | NW | PE), %eax
+	orl	$({cd} | {nw} | {pe}), %eax
```
daym, line 192: About the names inside the asm environment, do those have to be lower case (or is it convention to do that)?

I'd rather not have a constant be known under multiple different names, if it can be avoided.

Instead of `{cd}`, could it just be `{X86_CR0_CD}` like the rust constant is named anyway?

daym (changes requested):
About `const_asm` name, I can see it's difficult to find a name for that. Maybe something like `arch::global_asm_in_env` ?

Or just `arch::global_asm`, if that isn't too confusing (but maybe not...)

src/arch/x86/x86_64/src/bootblock.S
rminnich: or even have the crate name in there somehow? I have no idea how this works.

src/arch/x86/x86_64/src/lib.rs
```diff
@@ -3,13 +3,44 @@
...
 pub mod bzimage;
 pub mod consts;
 pub mod ioport;
 
+#[macro_export]
+macro_rules! const_asm {
+    ($asm:expr, $(options($($option:ident),*))? $(,)?) => {
+        global_asm!(
+            $asm,
+            pse = const $crate::consts::x86::X86_CR4_PSE,
```
rminnich, line 17: would calling it CR4_PSE help? Danny?

src/arch/x86/x86_64/src/bootblock.S
daym: global_asm It works kinda like (named) parameters in format strings (see .net, Python).

So you can have: `"hello {foo}, I heard {foo}".format(foo=2)` => `"hello 2, I heard 2"`. I don't think that having a `::` in the name of the placeholder is possible (would conflict with actual formatting specifier)--but if so, that would be nice indeed. (what is possible is to further enter into structs etc)

The assembly that will eventually become stable https://github.com/Amanieu/rfcs/blob/inline-asm/text/0000-inline-asm.md also supports something similar to this pull request in this regard.

src/arch/x86/x86_64/src/lib.rs
daym: I would be in favor of line 17 basename being exactly the same on both sides.

But in the definition of the constant itself (elsewhere), the `X86` in `x86::X86_CR4_PSE` is redundant anyway--so, sure, why not rename the latter too.

But really, I want to stress that this PR is pretty good as-is and if @retrage wants to do this stuff later or not at all, that's fine, too.

I've just been at the receiving end of one too many pointless rename-the-imports--where everyone then has to spend time mapping what is what using grep manually before one understands the source code.

src/arch/x86/x86_64/src/bootblock_nomem.S
```diff
@@ -21,15 +21,19 @@
  *
  */
 
+/* XXX: This list required to allow unused named arguments for rustc
+ * {pse} {pge} {pae} {efer} {lme} {cd} {nw} {ts} {mp} {pg} {wp} {pe}
+ * {pte_p} {pte_rw} {pte_ps} {pte2_mpat}
+ * {mtrr_cap_msr} {mtrr_def_type_msr} {mtrr_type_wrback} {mtrr_phys_mask_valid}
+ * {mtrr_def_type_en} {mtrr_type_wrprot}
+ */
```
daym, line 29: Hmm, it's kinda annoying that this is necessary. Isn't there an option for `global_asm` to turn this off maybe?

If it is necessary, we can make it official by just writing in the above that `the official interface is this: {pse} {pge} ` and so on.

jhand2: Looks like the reason this is needed is because of the `const_asm` macro. For every usage, it's injecting every variable, even if that specific asm file doesn't require it.

It seems fine to:
1. Export constants from the arch crates
2. Use the constants that are required for each `global_asm` invocation as-needed rather than having a unified macro that injects the environment. This seems more readable to me anyway, even if it's more verbose.

daym: Indeed, just not having the macro `const_asm` at all would also be a good solution--and just being explicit in what is needed instead, directly using `global_asm`, only with the bindings needed.

The only downside would be that you don't know what the available things in the `.S` file are while you have the `.S` file open in an editor. But those `.S` files hopefully won't need a lot of changes over time anyway--and even if we had the weird comment it would just be a workaround anyway. (in that case, it would still be better to just `include_str` a file which has the correct `global_asm` invocation AND the assembly source code in there)

To summarize, I agree that having `const_asm` automatically bind stuff internally is kinda weird.

retrage:
Thank you for the comments. I pushed some changes based on the feedbacks:
1. Use uppercase for named arguments b9f297f (I misunderstood the naming convention in Rust, uppercases in named arguments are allowed.)
2. Fix constant names so that they have the same name with named arguments as much as possible. 76bb132
3. Remove `const_asm` macro. I added this macro to reduce code redundancy, but it seems to be acceptable to have the same code in a different mainboard code.

jhand2 (dismissed):
This is really nice, much less nebulous than trying to do C preprocessing with crazy rust macros 😄

daym (dismissed):

retrage:
Thanks. I will clean up and rebase the commits before being merged.

daym (approved):

jhand2:
ship it!

PR #514, opened 2021-11-23
Dump HOB for qemu32, clean up warnings in said code from qemu

aquadara:

### DIFF
--- a/src/mainboard/emulation/qemu-fsp/src/main.rs
+++ b/src/mainboard/emulation/qemu-fsp/src/main.rs
@@ -120,55 +120,7 @@ pub extern "C" fn _start(_fdt_address: usize) -> ! {
     // "The bootloader should not expect a complete HOB list after the FSP returns
     // from this API. It is recommended for the bootloader to save this HobListPtr
     // returned from this API and parse the full HOB list after the FspSiliconInit() API."
-    unsafe {
-        write!(w, "EFI_HOB_HANDOFF_INFO_TABLE\r\n").unwrap();
-        write!(w, "========================================\r\n").unwrap();
-        write!(w, "Version = {}\r\n", (*hob_list_ptr).Version).unwrap();
-        write!(w, "BootMode = {}\r\n", (*hob_list_ptr).BootMode).unwrap();
-        write!(w, "EfiMemoryTop = {:#x?}\r\n", (*hob_list_ptr).EfiMemoryTop).unwrap();
-        write!(
-            w,
-            "EfiMemoryBottom = {:#x?}\r\n",
-            (*hob_list_ptr).EfiMemoryBottom
-        )
-        .unwrap();
-        write!(
-            w,
-            "EfiFreeMemoryTop = {:#x?}\r\n",
-            (*hob_list_ptr).EfiFreeMemoryTop
-        )
-        .unwrap();
-        write!(
-            w,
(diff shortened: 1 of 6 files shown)


aquadara:
The idea when I started this was going to be 1) add code to walk and dump the HOB in qemu32's main.rs, then 2) add generic code to fsp_common to support upstreaming said code from both the 32 and 64 bit code bases. Unfortunately, I don't think it's worth doing the latter after reading a bit about Rust generics. Most of it's described here https://stackoverflow.com/questions/64965194/access-field-of-a-generic-type-in-function, but the upshot is there would likely be more boilerplate code defining traits or a macro than code saved by upstreaming this.

jhand2:
One refactoring option here to make this a little cleaner:

FSP structures are defined per-CPU rather than per-board. So instead of unifying in fsp_common, I would recommend adding a function (like fsp_dump_hobs) to each CPU (src/vendorcode/fsp). Then at least different mainboards with the same CPU generation can rely this function.

src/mainboard/emulation/qemu-fsp/src/main.rs
```diff
@@ -110,45 +110,50 @@ pub extern "C" fn _start(_fdt_address: usize) -> ! {
     // from this API. It is recommended for the bootloader to save this HobListPtr
     // returned from this API and parse the full HOB list after the FspSiliconInit() API."
     unsafe {
```
jhand2, line 112: Which part of this is actually unsafe? It seems like it's mainly just dereferencing hob_list_ptr. It would be good to put just that part in an unsafe block and pull the rest out.

For example you can just dereference the pointer once in an unsafe block and then safely use that variable elsewhere.

src/mainboard/emulation/qemu-fsp/src/main.rs
```diff
@@ -110,45 +110,50 @@ pub extern "C" fn _start(_fdt_address: usize) -> ! {
...
+            write!(w, "Header.HobType = {}\r\n", (*hob_list_ptr).Header.HobType).unwrap();
             write!(
                 w,
                 "Header.HobLength = {}\r\n",
                 (*hob_list_ptr).Header.HobLength
-            );
+            )
+            .unwrap();
             hob_list_bytes_offset += (*hob_list_ptr).Header.HobLength as isize;
             hob_list_ptr = core::mem::transmute(hob_list_bytes_ptr.offset(hob_list_bytes_offset));
```
jhand2, line 158: Same here, the only part that's unsafe is just caclulating the next hob.

src/mainboard/emulation/qemu-fsp/src/main.rs
aquadara: Cleaned up while usptreaming, thanks! As suggested only the deferences are marked unsafe.

src/mainboard/emulation/qemu-fsp/src/main.rs
aquadara: Done!

aquadara:
> One refactoring option here to make this a little cleaner:
> 
> FSP structures are defined per-CPU rather than per-board. So instead of unifying in fsp_common, I would recommend adding a function (like fsp_dump_hobs) to each CPU (src/vendorcode/fsp). Then at least different mainboards with the same CPU generation can rely this function.

Done!

rjoleary (approved):

rjoleary (approved):

PR #520, opened 2021-12-10
Update oreboot to latest version

rjoleary:


orangecms:
We can go multiple steps. I have #525 now for bumping to 1.58.0, which requires a few things to resolve.

For a December release, we'd get 1.59.0 already, which seems to require linker adjustments, see #522.

orangecms:
FWIW, this PR can be closed for now, I think.

PR #516, opened 2021-12-03
SPI address line console

rminnich:
uart::spi::SPI implements a uart in which the characters are
output on SPI address lines 13:6. A properly patched em100
program can read this console. This is the console you use when things
are really screwed.

Doesn't even compile.

Signed-off-by: Ronald G. Minnich <rminnich@gmail.com>

PR #521, opened 2021-12-17
Load payload using FDT in qemu-armv7

derpsteb:
Beware: In its current state the PR changes `rust-toolchain.toml`.

This PR adds the function `get_kernel_area(..)` in order to retrieve a payload area object from the device tree located in flash.
It mostly removes the previously used `include_bytes` macro. The only usage left is to calculate the size of the payload for the final payload object.
To show the function working I updated the armv7 mainboard to use it. The mainboard also read the device tree from flash, based on a given address.

The PR also adds a tiny binary that prints a string to UART that I use for debugging. I thought it might be used for other people. I am not sure if we want this or if its preferable to remove binaries as much as possible.


Rwantare:
I think this should pass ci if you run `make format`

derpsteb:
I tried to do that, but it doesn't resolve the proc_macro problem.
Will keep the version with the updated nightly version for now.

orangecms:
`proc_macro2` was resolved via the toolchain upgrade, worked for #516 as well :)

orangecms:
Regarding the uart print binary:
There is a similar one already, `payloads/src/external/simple/`. It prints `testtesttest` to a fixed address `0x10010000` and is written in x86 assembly. Same thing for ARM - all good IMHO. :-)

PR #523, opened 2021-12-18
Draft: Test stable during CI

derpsteb:
This PR tries to run our normal CI jobs but using a recent stable toolchain. 

The new workflow is `build_stable`, which is a copy of `build`. So it runs LintAndFormat, Test and Build, but forces a stable toolchain for all these steps by running `sed` to change `rust-toolchain.toml`. 
For all steps `continue-on-error` is set to true, meaning that the pipeline won't fail if the job fails. Nonetheless, it allows us to see what breaks stable.


rjoleary (commented):
Please squash the commits

derpsteb:
I think we don't really need to merge this rn, as it will only slow down CI and not provide much value until someone tries to get stable in.

PR #526, opened 2021-12-20
Remove unneeded mainboards string

derpsteb:
So maybe this was just confusing to me but this is how I understand the mainboard(s) target atm:
```
MAINBOARDS := $(wildcard src/mainboard/*/*/Makefile)
# ...
.PHONY: mainboards $(MAINBOARDS)
mainboard: $(MAINBOARDS)
$(MAINBOARDS):
	cd $(dir $@) && make cibuild
``` 

Even though `mainboards` (mind the s) is declared as a phony target, there is no such target. **EDIT**: _There actually is, but it is empty._
However, there is a list of targets defined through the `$(MAINBOARDS)` variable. These _are_ phony and _exist_.
The `mainboards` variable is expanded to all of the makefiles that are found at the described path (first line), so its value is something like:
`src/mainboard/aaeon/upxtreme/Makefile src/mainboard/emulation/qemu-aarch64/Makefile src/mainboard/emulation/qemu-q35/Makefile ...`

Since all of these paths are targets and are declared as phony they will always be executed when they are called. Since `mainboard` (no s) depends on `$(MAINBOARDS)` it will also always execute, irrespective of the files currently present.

Originally I was thrown off by the fact that `make mainboards` wouldn't do anything, even though I thought there was a target for it (I missunderstood the phony keyword). I feel like it would be easier to understand if that was removed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,6 +37,6 @@ jobs:
     - name: 'Install Dependencies'
       run: sudo apt-get update && make ciprepare
     - name: 'Build all mainboards'
-      run: make --keep-going mainboard
+      run: make --keep-going mainboards
     - name: 'Generate report of binary sizes'
       run: ./scripts/generate-size-report.sh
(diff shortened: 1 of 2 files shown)


Rwantare:
Makes sense, update the README since we're dropping the s

derpsteb:
I started editing the readme, but after thinking about it I feel like the better approach is to rename the target to be plural. Forces the ci build stage to be run by default as well.

Rwantare (approved):

PR #531, opened 2022-01-07
Load payload using FDT in qemu-armv7

derpsteb:
Third time is the charm....

This PR adds the function get_kernel_area(..) in order to retrieve a payload area object from the device tree located in flash.
It mostly removes the previously used include_bytes macro. The only usage left is to calculate the size of the payload for the final payload object.
To show the function working I updated the armv7 mainboard to use it. The mainboard also read the device tree from flash, based on a given address.

The PR also adds a tiny binary that prints a string to UART that I use for debugging. I thought it might be used for other people. I am not sure if we want this or if its preferable to remove binaries as much as possible.


src/lib/device_tree/src/area.rs
```diff
@@ -15,6 +16,35 @@ pub struct Area {
...
 }
 
+pub fn get_kernel_area(dtfs_base: usize, dtfs_size: usize) -> Area {
+    let dtb = SectionReader::new(&Memory {}, dtfs_base, dtfs_size);
+    let areas = read_areas(&dtb).unwrap();
+    let mut last_area_end = 0;
+    let mut area_iter = areas.into_iter();
+
+    // Look for the first rompayload area and check for overlapping areas on the way
+    let area_opt = loop {
```
daym, line 26: Maybe add a comment that layoutflash ensures that fdt is sorted--otherwise the overlap check would be incomplete.

daym (approved):

src/lib/device_tree/src/area.rs
derpsteb: Done.

daym (approved):

PR #534, opened 2022-01-10
cargo: rename .cargo/config to .cargo/config.toml

luojia65:
This pull request renames all `.cargo/config` files into `.cargo/config.toml` files.

Cargo can now both identify files named 'config' or 'config.toml' under '.cargo' directory; these two file names have the same effect on compilation, but the 'toml' suffix make the files more readable (syntex highlights, etc.) under IDEs or text editors.


orangecms (approved):
Great idea, thank you!

PR #535, opened 2022-01-14
Initialize HPET driver in DebugPort

### LINKED ISSUE #500: qemu-q35 debug port print breaks boot
When running `make run` for qemu-q35 I can not get past the first print onto the debug port [here](https://github.com/oreboot/oreboot/blob/main/src/mainboard/emulation/qemu-q35/src/main.rs#L30-L32). 

When this line is commented out I can at least get up to loading the payload (L47), at which a panic will occur and the following message will be printed forever:
```
Welcome to oreboot                                                                                       
boot flag is ffff, not aa55, at ffff,                                                                                                                                                                              
                                                                                                                                                                                                                   
PANIC: Welcome to oreboot                                                                                                                                                                                          
boot flag is ffff, not aa55, at ffff,                                                                                                                                                                              
                                                                                                         
PANIC: Welcome to oreboot                                                                                                                                                                                          
boot flag is ffff, not aa55, at ffff,
```

Hopefully I will be able to resolve it myself, but maybe someone knows whats wrong here :)

derpsteb:
Add a function `enable()` to the HPET driver. This allows the calling driver, `DebugPort` in this case, enable the timer if it wants to use it. This is necessary for the qemu-q35 board as it does not init the timer, like the romecrb board for example.
This fixes #500.

I was contemplating whether `hpet0()` should do this automatically, but as the code is already used in other boards and there is a comment saying the caller should ensure enablement I felt like this is sensible solution. Happy to hear other opinions.

### DIFF
--- a/src/drivers/timer/src/hpet.rs
+++ b/src/drivers/timer/src/hpet.rs
@@ -13,6 +13,7 @@
 
 use core::ops;
 
+use tock_registers::interfaces::ReadWriteable;
 use tock_registers::interfaces::Readable;
 use tock_registers::register_bitfields;
 use tock_registers::registers::{ReadOnly, ReadWrite};
@@ -80,6 +81,16 @@ impl HPET {
     /// Returns a pointer to the register block
     fn ptr(&self) -> *const RegisterBlock {
         self.base as *const _
+    }
+
+    pub fn enable(&self) -> Result<(), &str> {
+        // set enable bit for HPET timer as per 2.3.5
+        // https://www.intel.com/content/dam/www/public/us/en/documents/technical-specifications/software-developers-hpet-spec-1-0a.pdf
+        self.config.modify(CONFIG::TMREN::SET);
+        match self.config.is_set(CONFIG::TMREN) {
+            true => Ok(()),
+            false => Err("HPET.enable() failed to set timer_enable."),
+        }
     }
 
     // Sleeps at least the given amount of time (in fs).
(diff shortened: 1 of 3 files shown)


src/drivers/timer/src/hpet.rs
```diff
@@ -82,6 +83,16 @@ impl HPET {
         self.base as *const _
     }
 
+    pub fn enable(&self) -> Result<(), &str> {
+        // set enable bit for HPET timer as per 2.3.5
+        // https://www.intel.com/content/dam/www/public/us/en/documents/technical-specifications/software-developers-hpet-spec-1-0a.pdf
+        self.config.set(1);
```
Rwantare, line 89: Instead of doing a raw set, you can use `config.modify(CONFIG::TMREN::SET)` since Danny already defined the type on L63

derpsteb: I have to take a closer look at this tock_registers crate. Looks great, thanks!

derpsteb: Resolved. Danny showed me another trick, which is great.

src/drivers/uart/src/debug_port.rs
```diff
@@ -20,8 +20,8 @@ impl<D: Driver> DebugPort<D> {
 }
 
 impl<D: Driver> Driver for DebugPort<D> {
-    // Nothing to set up here
     fn init(&mut self) -> Result<()> {
+        self.timer.enable().unwrap();
```
daym: Probably makes more sense to `map_err` here. `unwrap` would just fail the program on error.

rminnich (approved):

daym (dismissed):

rminnich:
you should be able to run 
make format
locally from top of tree to catch formatting issues

derpsteb:
yeah I still have to develop the habit :(

daym (approved):

PR #537, opened 2022-01-15
set coverall

gogo2464:
Signed-off-by: gogo <gogo246475@gmail.com>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,9 +26,14 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: 'Install Dependencies'
-      run: sudo apt-get update && make ciprepare
-    - name: 'Run rust tests'
-      run: make --keep-going test
+      run: |
+        sudo apt-get update && make ciprepare
+    - name: 'Create coverage file'
+      run: make --keep-going citest
+    - name: 'upload the coverage file to server'
+      uses: codecov/codecov-action@v2.1.0
+      with:
+        files: ./coverall/lcov.txt
 
   Build:
     runs-on: ubuntu-latest
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,7 @@ firsttime:
 	rustup target add riscv64gc-unknown-none-elf
 	rustup target add aarch64-unknown-none-softfloat
 	rustup target add riscv32imc-unknown-none-elf
+	cargo install cargo-tarpaulin
 
 nexttime:
 	rustup toolchain install $(TOOLCHAIN_VER)
@@ -107,9 +108,24 @@ checkformat: $(CRATES_TO_CHECKFORMAT)
 # Once those are fixed, we can just use a test target.
 CRATES_TO_TEST := $(patsubst %/Makefile,%/Makefile.test,$(ALLMAKEFILE))
 $(CRATES_TO_TEST):
-	cd $(dir $@) && make citest
+	cd $(dir $@) && make test
 .PHONY: test $(CRATES_TO_TEST)
-test: $(CRATES_TO_TEST)
+
+# NOTE: In CI, we run tests with coverage report.
+# The individual crates' Makefiles use the `citest` target.
+# However, there are a number of crates which can not test.
+# Hence, `citest` either points to `coverage` or `skiptest`.
+# We use the LCOV format so that we can simply concatenate
+# the multiple reports. See ./Makefile.inc for details.
+CRATES_TO_CITEST := $(patsubst %/Makefile,%/Makefile.citest,$(ALLMAKEFILE))
+$(CRATES_TO_CITEST):
+	cd $(dir $@) && make citest
+.PHONY: test $(CRATES_TO_CITEST)
+citest: $(CRATES_TO_CITEST)
+	# concatenate all the results from the indidividual directories
+	mkdir -p coverall
+	find . -name "lcov.info" -exec cat > coverall/lcov.txt {} +
+
 
 # TODO: Remove write_with_newline
 CRATES_TO_CLIPPY := $(patsubst %/Makefile,%/Makefile.clippy,$(ALLMAKEFILE))
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -37,7 +37,7 @@ checkformat:
 	cargo fmt -- --check
 
 clean:
-	rm -rf target/
+	rm -rf target/ coverall/
 
 # Clippy is the standard rule
 clippy:
@@ -54,6 +54,10 @@ skipclippy:
 
 # Testing. A lot of directories can't test either, yet.
 test:
+	cargo test
 
 skiptest:
 	echo Not doing test for this target
+
+coverage:
+	cargo tarpaulin --out Lcov
--- a/payloads/Makefile
+++ b/payloads/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../)
 include ../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
--- a/src/arch/ppc64/Makefile
+++ b/src/arch/ppc64/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../)
 include ../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ oreboot.asm
 #
 #already existing elements are commented out
 
+coverall
 target
 **/*.rs.bk
 debug.log
--- a/src/arch/arm64/armv8/Makefile
+++ b/src/arch/arm64/armv8/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../../)
 include ../../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
--- a/src/arch/x86/x86_64/Makefile
+++ b/src/arch/x86/x86_64/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../../)
 include ../../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: skiptest
--- a/src/console/Makefile
+++ b/src/console/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../)
 include ../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
--- a/src/cpu/amd/Makefile
+++ b/src/cpu/amd/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../)
 include ../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
--- a/src/cpu/armltd/cortex-a53/Makefile
+++ b/src/cpu/armltd/cortex-a53/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../../)
 include ../../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
--- a/src/drivers/clock/Makefile
+++ b/src/drivers/clock/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../)
 include ../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
--- a/src/drivers/model/Makefile
+++ b/src/drivers/model/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../)
 include ../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
--- a/src/drivers/sifive/spi/Makefile
+++ b/src/drivers/sifive/spi/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../../)
 include ../../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
--- a/src/drivers/timer/Makefile
+++ b/src/drivers/timer/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../)
 include ../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
--- a/src/drivers/uart/Makefile
+++ b/src/drivers/uart/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../)
 include ../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
--- a/src/drivers/wrappers/Makefile
+++ b/src/drivers/wrappers/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../)
 include ../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
--- a/src/lib/consts/Makefile
+++ b/src/lib/consts/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../)
 include ../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
--- a/src/lib/device_tree/Makefile
+++ b/src/lib/device_tree/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../)
 include ../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
--- a/src/lib/print/Makefile
+++ b/src/lib/print/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../)
 include ../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
--- a/src/lib/uefi/Makefile
+++ b/src/lib/uefi/Makefile
@@ -1,4 +1,5 @@
 OREBOOT=$(abspath $(CURDIR)/../../../)
 include ../../../Makefile.inc
 ciclippy: clippy
-citest: test
+#todo bring back tests
+citest: skiptest
--- a/src/lib/util/Makefile
+++ b/src/lib/util/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../)
 include ../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
--- a/src/mainboard/sunxi/nezha/sbi/Makefile
+++ b/src/mainboard/sunxi/nezha/sbi/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../../../)
 include ../../../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: skiptest
--- a/src/soc/amd/common/boot/Makefile
+++ b/src/soc/amd/common/boot/Makefile
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../../../../../)
 include ../../../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: coverage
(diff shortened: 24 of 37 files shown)


gogo2464:
just a tmp PR to debug.

gogo2464:
Coverage works but the test already used to fail.

Makefile.inc
```diff
@@ -54,6 +54,9 @@ skipclippy:
 
 # Testing. A lot of directories can't test either, yet.
 test:
+	cargo tarpaulin --ciserver travis-ci --coveralls ${{ secrets.GITHUB_TOKEN }} ;
+	cargo tarpaulin --out Xml ;
+	curl -s "https://codecov.io/bash" | bash -s
```
orangecms, line 59: Can we use this here instead from the official docs?

https://github.com/marketplace/actions/codecov#usage

Coming from https://about.codecov.io/tool/github-actions/

gogo2464: Instead of Tarpaulin? It is also in the official doc: https://about.codecov.io/tool/tarpaulin/. Your link is about python. I am a bit confused.

orangecms: Sorry, I meant using the Action from the Marketplace instead of executing a foreign shell script.
Does that work for Tarpaulin? I've never used Tarpaulin, to be honest.

gogo2464: Ah ok! Then is `curl -s "https://codecov.io/bash" | bash -s` only the single embarassing line? I think I understood you want I rewrite the script from bash file to the Makefile. If it is that I can do it.

Yes it works for Tarpaulin. I just need to ignore some tests.

orangecms: Yea I just meant the Action. Let's give it a shot! :)

gogo2464:
the coverage stops due to an error. See https://github.com/oreboot/oreboot/runs/4872023670?check_suite_focus=true#step:4:7799 But I do not find any log for this error. Where could I find it please?

orangecms:
> the coverage stops due to an error. See https://github.com/oreboot/oreboot/runs/4872023670?check_suite_focus=true#step:4:7799 But I do not find any log for this error. Where could I find it please?

I see two things:

```
Jan 19 18:32:29.874  INFO cargo_tarpaulin::process_handling: running /home/runner/work/oreboot/oreboot/src/lib/uefi/target/debug/deps/tests-7d92034841aed921

running 1 test
test parse_fsp ... FAILED
```
and this is a bit weird
```
Compiling wrappers v0.1.0 (/home/runner/work/oreboot/oreboot/src/drivers/wrappers)
error: Broken pipe (os error 32)
warning: build failed, waiting for other jobs to finish...
error: Broken pipe (os error 32)
error: build failed
Jan 19 18:36:38.554 ERROR cargo_tarpaulin: Failed to compile tests! Error: arch: linking with `cc` failed: exit status: 1
Error: "Failed to compile tests! Error: arch: linking with `cc` failed: exit status: 1"
make[1]: *** [../../../../Makefile.inc:57: test] Error 1
make[1]: Target 'citest' not remade because of errors.
make[1]: Leaving directory
```

.github/workflows/build.yml
```diff
@@ -25,10 +25,15 @@ jobs:
...
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 2
     - name: 'Install Dependencies'
       run: sudo apt-get update && make ciprepare
     - name: 'Run rust tests'
-      run: make --keep-going test
+      run: |
+        sudo apt install libssl-dev;
+        cargo install cargo-tarpaulin;
```
orangecms, line 35: The installation should be part of the `'Install Dependencies'` step.

gogo2464: I can only agree with you but I need to make it work first. It does not work when I change the places.

orangecms: Sorry, that was only half of it anyway. The `libssel-dev` headers should be in that step. The cargo stuff should be part of the `ciprepare` target and wrapped in make; see `Makefile` for how it's done with other tools.

Also wrap the invocations in `make` in general. Let's see if that works then. :-)

gogo2464:
@orangecms I do not see such errors. Even with ctr-f. Are you sure it is not the warn `WARN cargo_tarpaulin::report::coveralls: Failed to collect git info: failed to get branch name: reference 'HEAD' is neither a local nor a remote branch.; class=Invalid (3) ` that blocks the tests?

orangecms:
> @orangecms I do not see such errors. Even with ctr-f. Are you sure it is not the warn `WARN cargo_tarpaulin::report::coveralls: Failed to collect git info: failed to get branch name: reference 'HEAD' is neither a local nor a remote branch.; class=Invalid (3) ` that blocks the tests?

The default colored view is a bit hard to search in. It works better in the raw view: https://pipelines.actions.githubusercontent.com/Ug5IXs0NP0Hd2f59Ys1GY2AxeChWsITKXSL8dQKC6LMA4hU13h/_apis/pipelines/1/runs/715/signedlogcontent/6?urlExpires=2022-01-20T01%3A34%3A07.5885730Z&urlSigningMethod=HMACV1&urlSignature=y%2FNpcpiZYZXthV5u5SQYIIsZlpdepmsq%2BPO3jseVwEA%3D

Not sure if the link here works; it's an option behind the three dots `...` menu.

gogo2464:
No. The link does not works.

orangecms:
> No. The link does not works.

Then please follow the link in the option menu.

gogo2464:
Uri expired

gogo2464:
It passes any test :D. Please do not merge. I skipped too much tests. I have to put some tests back and if possible solve the issue on the last test.

.github/workflows/build.yml
```diff
@@ -25,10 +25,15 @@ jobs:
...
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 2
     - name: 'Install Dependencies'
       run: sudo apt-get update && make ciprepare
     - name: 'Run rust tests'
-      run: make --keep-going test
+      run: |
+        sudo apt-get install libssl-dev;
```
rjoleary, line 34: Remove semicolons from the end of the lines.

Makefile.inc
gogo2464: According to the official doc https://docs.travis-ci.com/user/coveralls/#using-coveralls-with-other-languages there is no doc for coverall on rust. :/ I am digging.

orangecms:
Nice, getting there! :-)
Some more feedback and suggestions:

Could you move the `apt install libssl-dev` into the `debiansysprepare` target in the `Makefile`?
There is a helper named `CARGOINST` to wrap the `cargo install cargo-tarpaulin` using our toolchain. That goes into `firsttime` in `Makefile`.
To be able to run the tests locally without the coverage overhead, there could be a specific target for it, let's say just `coverage`.

With the above, everything is more self-contained and the install step in the CI yml can stay as it is. And finally, we could run tests with coverage locally in the individual directories to iterate over it without waiting for CI.

What do you think?

orangecms:
Addendum: Itt looks to me like it's fine to ignore some tests now. We are currently skipping all of them, aren't we? So there should be a `cargo test` in the `test` target in `Makefile.inc` and a `coverage` target with the tarpaulin stuff for CI, right?

Makefile
```diff
@@ -51,6 +51,7 @@ firsttime_fsp:
 
 debiansysprepare:
 	sudo apt-get install device-tree-compiler pkg-config libssl-dev llvm-dev libclang-dev clang qemu-system-x86
+	sudo apt-get install libssl-dev
 	# --default-toolchain is purely an optimization to avoid downloading stable Rust first.
```
orangecms, line 55: Ah, I just saw that this is already installed in the line above, so this line can be removed as it's redundant.

gogo2464:
Sounds ready to merge. Do you have access to the coverall online? I do not see the bot anymore.

.github/workflows/build.yml
```diff
@@ -25,10 +25,22 @@ jobs:
...
-      run: sudo apt-get update && make ciprepare
+      run: |
+        sudo apt-get update && make ciprepare
+        sudo apt-get install libssl-dev
+        cargo install cargo-tarpaulin
     - name: 'Run rust tests'
-      run: make --keep-going test
+      run: |
+        make --keep-going test &&
+        cargo tarpaulin --out Xml --output-dir ./coverall
```
orangecms, line 38: this should now be `make --keep-going coverage`

Makefile.inc
```diff
@@ -54,6 +54,11 @@ skipclippy:
...
 # Testing. A lot of directories can't test either, yet.
 test:
+	cargo test
 
 skiptest:
 	echo Not doing test for this target
+
+coverage:
+	cargo tarpaulin --out Xml --output-dir ./coverall
+	curl -s "https://codecov.io/bash" | bash -s
```
orangecms, line 64: remove the curl here, that's covered by the GitHub Action and does not work anywhere else

gogo2464: I wanted to allow people to submit coverage from their own computer. Do I do another make recipe?

gogo2464: I wanted to allow people to submit coverage from their own computer. Do I do another make recipe?

.github/workflows/build.yml
orangecms: note: _only_ run the coverage, as it implies running the tests already

.github/workflows/build.yml
```diff
@@ -25,10 +25,22 @@ jobs:
...
+    - name: 'upload the xml to coverall server'
+      uses: codecov/codecov-action@v2.1.0
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
```
orangecms, line 42: This is not required for public repos as per the GitHub example at
https://crates.io/crates/cargo-tarpaulin

Makefile.inc
orangecms: When running the command locally, you also get the reports locally. However, due to access rights, there is no chance for arbitrary users pushing coverage to arbitrary projects. That is what the access controls implement between GitHub and other services. Otherwise, anyone could interfer with any project. Instead, if you wanted to publish your own, you could do so from your own fork and set up your own CI services for it, run your own infra, etc..

.github/workflows/build.yml
```diff
@@ -25,10 +25,22 @@ jobs:
...
-      run: sudo apt-get update && make ciprepare
+      run: |
+        sudo apt-get update && make ciprepare
+        sudo apt-get install libssl-dev
+        cargo install cargo-tarpaulin
     - name: 'Run rust tests'
-      run: make --keep-going test
+      run: |
+        make --keep-going test &&
+        make --keep-going converage
```
orangecms, line 38: Typo: co*n*verage
remove the first part, tarpaulin already runs the tests

orangecms:
> Addendum: It looks to me like it's fine to ignore some tests now. We are currently skipping all of them, aren't we? So there should be a `cargo test` in the `test` target in `Makefile.inc` and a `coverage` target with the tarpaulin stuff for CI, right?

Correction: See here for the actual entry point. That's where you'll need to add the top-level `coverage` target in the same fashion, and then call `make coverage` for each resolved directory:
https://github.com/oreboot/oreboot/blob/main/Makefile#L110

Then adjust the `citest` target in each Makefile. We will no longer need it.
The new convention will be: In CI, run every test that works (as of now), and do so with the `coverage` target.

Sorry for the bunch of work here, the build system has its complexity, you're tidying it up in the process. :-)  ✨ 🧹

gogo2464:
Sounds perfect. If you agree I can definitely merge these two commits and then merge immediately.

.github/workflows/build.yml
```diff
@@ -25,10 +25,19 @@ jobs:
...
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 2
     - name: 'Install Dependencies'
-      run: sudo apt-get update && make ciprepare
+      run: |
+        sudo apt-get update && make ciprepare
+        sudo apt-get install libssl-dev
+        cargo install cargo-tarpaulin
```
orangecms, line 34: this should be restored since the install happens via the make ciprepare target

orangecms:
> Sounds perfect. If you agree I can definitely merge these two commits and then merge immediately.

Hmm, please look through the previous discussions. A few things are open.

Makefile
```diff
@@ -130,3 +130,8 @@ clean:
 	rm -rf $(wildcard src/*/*/target) \
 	rm -rf $(wildcard src/*/*/*/target) \
 	rm -rf $(wildcard src/*/*/*/*/target)
+
+
+sendlocalcoverall:
+	cargo tarpaulin --out Xml --output-dir ./coverall
+	curl -s "https://codecov.io/bash" | bash -s
```
orangecms, line 137: This cannot be used locally. Please try it to see what happens. It can be removed.

Instead, define a `coverage` target, which is similar to the `test` target, but should run the `coverage` task in each of the directories.

gogo2464: I do not really see how you would like I use the 
`
uses: codecov/codecov-action@v2.1.0
`

in the Makefile.

orangecms: That is a different thing. The _sending_ of the report does not belong in `Makefile`. What goes here is _generating_ it, i.e., invoking `coverage` for the individual directories, and there needs to be a corresping target that calls tarpaulin in `Makefile.inc`.
Please look at the `test` target for reference.

Makefile
```diff
@@ -130,3 +130,8 @@ clean:
 	rm -rf $(wildcard src/*/*/target) \
 	rm -rf $(wildcard src/*/*/*/target) \
 	rm -rf $(wildcard src/*/*/*/*/target)
+
+
+coverage:
+	cd $(dir $@) && make cicoverage
+.PHONY: coverage
```
orangecms, line 137: This also needs the `$(CRATES_TO_TEST)` like the other targets above

gogo2464:
The coverage xml file is alway overwritten for each crate. I have to fix it.

.github/workflows/build.yml
orangecms: please restore this and do the cargo install in the Makefile - otherwise this causes the tests to run longer

.github/workflows/build.yml
```diff
@@ -25,10 +25,23 @@ jobs:
...
     - name: 'Run rust tests'
       run: make --keep-going test
+    - name: 'Generate coverall coverage'
+      run: make --keep-going coverage
+    - name: 'rename coverall file'
+      run: make movefile
+    - name: 'upload the xml coverage to coverall server'
```
orangecms, line 41: either remove "coverall" or change it into "codecov" - they are two different services

Makefile
```diff
@@ -130,3 +130,17 @@ clean:
...
+.PHONY: coverage $(CRATES_TO_COVER)
+
+$(CRATES_TO_COVER):
+	echo $(CRATES_TO_COVER)
+	cd $(dir $@) && make cicoverage folder=`basename $(dir $@)`
+
+.PHONY: coverage $(CRATES_TO_COVER)
+
+movefile:
+	mv ./coverall/lcov.temp ./coverall/lcov.txt
```
orangecms, line 146: this extra target is not necessary - just `cat` to `lcov.txt` directly

.github/workflows/build.yml
```diff
@@ -25,10 +25,23 @@ jobs:
...
+      uses: codecov/codecov-action@v2.1.0
+      with:
+        directory: ./coverall
```
orangecms, line 44: this should be `files:`, as per docs: https://github.com/codecov/codecov-action#usage

Makefile.inc
```diff
@@ -54,6 +54,15 @@ skipclippy:
...
 test:
+	cargo test
 
 skiptest:
 	echo Not doing test for this target
+
+coverage:
+	cargo tarpaulin --out Lcov --output-dir "$(OREBOOT)/coverall"
+	cat `echo "$(OREBOOT)/coverall/lcov.info"`.info >> $(OREBOOT)/coverall/lcov.temps
+	rm "$(OREBOOT)/coverall/lcov.info"
```
orangecms, line 65: this is not necessary; for cleanup, we can include this in the `clean` target (can be done later also)

Makefile.inc
```diff
@@ -54,6 +54,15 @@ skipclippy:
...
 # Testing. A lot of directories can't test either, yet.
 test:
+	cargo test
 
 skiptest:
 	echo Not doing test for this target
+
+coverage:
+	cargo tarpaulin --out Lcov --output-dir "$(OREBOOT)/coverall"
+	cat `echo "$(OREBOOT)/coverall/lcov.info"`.info >> $(OREBOOT)/coverall/lcov.temps
```
orangecms, line 64: This needs to be all the `lcov.info` files in all the `coverall` directories - you can use `find` and `exec` to achieve this

payloads/Makefile
```diff
@@ -2,3 +2,4 @@ OREBOOT=$(abspath $(CURDIR)/../)
 include ../Makefile.inc
 ciclippy: clippy
 citest: test
+cicoverage: coverage
```
orangecms, line 5: please keep it simple and expand on the `citest` target instead (in all the `Makefile`s)

src/drivers/uart/Cargo.lock
```diff
@@ -3,63 +3,23 @@
...
-version = "0.7.18"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
-dependencies = [
- "memchr",
-]
-
-[[package]]
-name = "atomic-polyfill"
+name = "as-slice"
```
orangecms, line 6: please restore the file, it's not the scope of test coverage

src/mainboard/ast/ast25x0/Cargo.lock
```diff
@@ -1,342 +0,0 @@
-# This file is automatically @generated by Cargo.
```
orangecms, line 1: please restore this file

Makefile.inc
```diff
@@ -54,6 +54,15 @@ skipclippy:
...
 
 skiptest:
 	echo Not doing test for this target
+
+coverage:
+	cargo tarpaulin --out Lcov --output-dir "$(OREBOOT)/coverall"
+	cat "$(OREBOOT)/coverall/lcov.info" >> $(OREBOOT)/coverall/lcov.temps
+	rm "$(OREBOOT)/coverall/lcov.info"
+
+skipcoverage:
```
orangecms, line 67: this is not necessary; either a test can run, then it can also be run with coverage, or skip it, using `skiptest`

orangecms (changes requested):
Please restore the deleted files and look at the individual remarks, thank you!

payloads/Makefile
gogo2464: Are you sure? I spent a lot of time to split the tasks. I believed it was better like that because now we can test with no coverage.

orangecms: We can already do this. That's what the simple `test` target is for. :)

payloads/Makefile
```diff
@@ -1,4 +1,4 @@
 OREBOOT=$(abspath $(CURDIR)/../)
 include ../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: test coverage
```
orangecms, line 4: only call `coverage`; as explained, it already _includes_ running the tests

src/arch/arm/armv7/Makefile
```diff
@@ -1,4 +1,5 @@
 OREBOOT=$(abspath $(CURDIR)/../../../../)
 include ../../../../Makefile.inc
 ciclippy: clippy
-citest: skiptest
+citest: skiptest skipcoverage coverage
```
orangecms, line 4: restore this; just `skiptest`

Makefile.inc
```diff
@@ -54,6 +55,14 @@ skipclippy:
 
 # Testing. A lot of directories can't test either, yet.
 test:
+	cargo test
 
 skiptest:
 	echo Not doing test for this target
+
+coverage:
+	cargo tarpaulin --out Lcov --output-dir "$(OREBOOT)/coverall"
```
orangecms, line 64: This still needs to be addressed. Crates should be self-contained, so remove the `--output-dir` and instead concatenate the resulting files for the main `Makefile` target `citest` after the individual reports were created.
Please also add `coverall/` to `.gitignore`.

Makefile.inc
orangecms: remove this

src/arch/arm/armv7/Makefile
```diff
@@ -1,4 +1,5 @@
 OREBOOT=$(abspath $(CURDIR)/../../../../)
 include ../../../../Makefile.inc
 ciclippy: clippy
-citest: skiptest
+citest: skiptest skipcoverage
```
orangecms, line 4: restore this, only run `skiptest`, also check the other files

src/arch/ppc64/Makefile
```diff
@@ -1,4 +1,5 @@
 OREBOOT=$(abspath $(CURDIR)/../../../)
 include ../../../Makefile.inc
 ciclippy: clippy
-citest: test
+citest: test coverage
```
orangecms, line 4: remove `test`, only run coverage, and please remove the extra empty line below

orangecms (dismissed):
:100: excellent, thank you so much!

orangecms (approved):

Issue #497, opened 2021-10-29
Assembly preprocessor (rpp) improvements


jhand2:
rpp does not currently support multiline macros such as
```
#define MACRO some \
    stuff
```

This should be an easy thing to add

pdogr:
Hi!
I've tried adding this feature in #538. I was not sure if we should preserve the exact structure (escaped newlines) in macro definition, though this would really depend on whether the assembly has a free-form syntax?

jhand2:
Thanks! It's worth noting that since https://github.com/oreboot/oreboot/issues/504 was resolved, oreboot assembly no longer uses rpp to insert macros. Apologies for not updating this issue.

That says, rpp is still fun endeavor and I'll take a look at the PR.

pdogr:
Sorry! I should've asked before picking this up.

jhand2:
No problem :smile: I might move rpp into it's own repo/crate at some point.

PR #538, opened 2022-01-25
feat: introduce multi line macro in rpp 

pdogr:
#497 
This PR attempts to adds multi-line macro support in rpp. Macros like the one given below are supported now. 
```
#define MACRO \
 (A + \
  B) // This expands to (A + B)
```
Changes introduced:
 - Add `split_lines_re` Regex to `Context` struct which is used to split a line preserving escaped newlines. 
 - Add test `macro_with_newline` which checks multi-line macro expansion. 
 - The current behavior is to ignore escaped newlines while macro expansion, so multi-line macros essentially are treated as a big single-line macro.

### DIFF
--- a/tools/rpp/src/lib.rs
+++ b/tools/rpp/src/lib.rs
@@ -37,6 +37,7 @@ pub struct Context {
     word_re: Regex,
     block_comment_re: Regex,
     eol_comment_re: Regex,
+    split_lines_re: Regex,
 }
 
 impl Context {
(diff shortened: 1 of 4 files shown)


gogo2464:
Your commit is not signed. Do git commit --ammend --signoff to fix it.

tools/rpp/src/lib.rs
```diff
@@ -172,8 +183,11 @@ pub fn process_str(s: &str, ctx: &mut Context) -> Result<String, ParsingError> {
     });
 
     let mut out = String::new();
-    for line in no_comments.lines() {
-        ctx.line_num += 1;
+
+    // Collect range of splits
+    let split_lines_range: Vec<_> = ctx.split_lines_re.find_iter(&no_comments).collect();
```
jhand2, line 188: What if you just did a find/replace of line continuations (lines ending in `\`) and then do the same as what's done for block comments (above). Then you shouldn't need any of the other changes in this file.

pdogr: This was used so as to keep track of the correct line number. If lines ending in `\` are escaped here, wouldn't that mess up the line number in process_line?

rjoleary (approved):

PR #544, opened 2022-02-07
fix UEFI lib

### LINKED ISSUE #542: broken src/lib/uefi parsing and tests
1. The crate `src/lib/uefi` broke in `5beeb5641124c240a06bb889605e2f44485cbf26`, where the loops were removed that iterate over the FV (firmware volume).
2. Its tests broke in `b8f0e5100abbdbd396ac884e827bce5e2cabb4c0` where the `fsp` submodule was updated, but the test cases depend on the binaries in it (12 of them).

The first issue is easy to fix, but the test setup is fragile.

Question: What should we prefer, a second fsp submodule side by side with the uefi lib, pinned to its version for testing, or checking the respective files (some MB involved) into the repo (maybe reduce them? 12 feels like it's not really providing value), or... any other ideas?

orangecms:
this resolves #542

### DIFF
--- a/src/lib/uefi/src/lib.rs
+++ b/src/lib/uefi/src/lib.rs
@@ -209,11 +209,10 @@ where
         }
 
         // Iterate through files.
-        let res = {
+        while {
             index = (index + 7) & !7; // align to 8 bytes
             index < fv_end
-        };
-        if res {
+        } {
             let ffs_base = index;
 
             // Check if the header is empty.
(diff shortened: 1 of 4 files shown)


src/lib/uefi/src/lib.rs
```diff
@@ -209,11 +209,10 @@ where
         }
 
         // Iterate through files.
-        let res = {
+        while {
```
gogo2464, line 212: Why do you change this. Did this use to work?

gogo2464:
Sounds fine. It does not break the tests and the fsp-bin name sounds better.

src/lib/uefi/src/lib.rs
gogo2464: It probably use to break tests. I remember.

gogo2464 (approved):

src/lib/uefi/src/lib.rs
orangecms: Yes, it's restored from the original implementation, see the commit message.

Issue #443, opened 2021-06-26
Re-enable RISC-V QEMU Tests in CI


rjoleary:
See the `exit 0` skip here: https://github.com/oreboot/oreboot/blob/main/azure-pipelines.yml#L133

These CI checks were disabled because they re-build QEMU from source each time which added an unnecessary 8 minutes to the tests. Instead, we should be using a pre-built QEMU or caching the build.

derpsteb:
Could use [this](https://github.com/actions/cache) caching action. 
Would requiring adding another step to the job. Then the other jobs could check if there is a cache hit before running.
Initial ideas for a cache-key could be: The built binary or the Makefile. Both only exist if at least one build has run. The Makefile potentially wouldn't catch version changes.

orangecms:
Note: We have switched to GitHub Actions; now see
https://github.com/oreboot/oreboot/blob/main/.github/workflows/fsp-qemu.yml#L72

PR #546, opened 2022-02-08
Get QEMU tests up and running

alistair23:
This fixes https://github.com/oreboot/oreboot/issues/443 by getting QEMU up and running for the virt board.

We also remove the QEMU specific workarounds. The current QEMU workarounds don't work, so let's just remove them. The HiFive1 is an outdated board, not very well modelled on QEMU and is not being worked on. Instead of leaving in some old hacks let's just remove then and remove it from the CI.


orangecms (approved):
Thank you for cleaning this up! :-)

PR #549, opened 2022-02-18
Add an ad-hoc tool for rewriting a "shrunk" DTFS based on PAYLOAD_A size

aquadara:


rminnich (approved):

src/mainboard/emulation/qemu-q35/Makefile
```diff
@@ -8,3 +8,20 @@ include ../../../../Makefile.mainboard.inc
 cibuild: $(IMAGE)
 ciclippy: skipclippy
 citest: skiptest
+
```
rminnich, line 11: Can you put this in the top-level Makefile.inc so all mainboards can use it.

rminnich: let's do that later, once this is tested, actually.

PR #552, opened 2022-02-26
Add a .cargo/config for trembyle

rminnich:


orangecms (changes requested):
There is already a `config.toml` - which we use by convention now, since it enables syntax highlighting to have the extension. It's the same file, except that it points to the existing `link.ld`.
Now Cargo prefers the config file without the extension for some reason (probably hardcoded order in Cargo?),  and then linking errors because `link-coreboot.ld` does not exist.
Did you mean to switch to that different ld file and forgot to commit it, or... is this a leftover from some earlier point in time?

PR #553, opened 2022-03-24
Rearrange oreboot driver modules

luojia65:
This pull request rearranges oreboot driver modules into one Cargo project. It also includes fixes on dependency for oreboot mainboards and SoCs to comply with this rearrangement.

It includes following commits:

1. Rearrange all drivers into one Cargo project
2. Adjust architecture support modules
3. Adapt mainboards and SoCs into new structure

As is discussed in OSF community, the current oreboot project matches the directory structure in C-based coreboot project, it results in every driver folder it includes a complete Cargo crate. It's too heavy for each directory and we can't share incremental compilation caches in one workspace. The current project structure is familiar for C coders, but it resulted in disk usage and existence of makefiles. Also, it would be hard to prepare developing environment, especially for new Rust contributors.

I'm working to rearrange the project, this pull request would be the first part. Except for the need of changing dependency paths, this PR does not modify working code on mainboards. This pull request does not yet include a reformation of build system.

I adapted the following mainboards to new structure and they may produce binaries successfully:

- aaeon/upsquared
- aaeon/upxtreme
- amd/romecrb
- asrock/a300m-stx
- ast/ast25x0
- emulation/qemu-aarch64
- emulation/qemu-armv7
- emulation/qemu-fsp
- emulation/qemu-fsp32
- emulation/qemu-q35
- emulation/qemu-riscv
- google/trembyle
- seeed/beaglev
- sifive/hifive
- sunxi/nezha

After this reformation, we may furtherly arrange the oreboot project under one Cargo workspace. It would share compilation caches, and a xtask-based build framework would be possible to introduce within one arranged workspace.


orangecms:
This is quite huge. :sweat_smile: Can we do this gradually and split things up a bit? Like, can this be done with each of the patches individually? That'd make the review easier, I think.

Regarding the directory names: Could you explain the reason for renaming drivers to oreboot-drivers? Can't we keep it as it is, just `drivers`?

Anyway, this is really cool, seeing how much we could simply drop, 1000 lines by juat rearranging - thank you a lot! :partying_face:

luojia65:
Hello! I'll explain as follows.

> Can we do this gradually and split things up a bit? Like, can this be done with each of the patches individually? That'd make the review easier, I think.

I initially submit this pull request as whole, for this modification affects depenency path of downstream mainboard crates. Commits submitted first modified project structure, it would result in failure of compilation for mainboard crates. The following commit fixes this problem by synchronizing structure in Cargo.toml dependency files. If this caused difficulty in review, I'd split this pull request into three, but first and second pull request won't build and pass CI checks.

> Regarding the directory names: Could you explain the reason for renaming drivers to oreboot-drivers? Can't we keep it as it is, just `drivers`?

Cargo crate names are what it represents in Rust source code and surrounding ecosystem. If we call it `drivers`, this name would be too general for Rust ecosystem projects to reference it in source code (we already did this in oreboot project, `tock-registers` is a register representation in `tock` project). Name `oreboot-drivers` would be a better name for a cargo crate and source code as it represents drivers that are specific to oreboot project, where other projects can make use of if necessary.

Thank you! If applicable, please inform me and I'll split this pull request into smaller parts.

orangecms:
> Hello! I'll explain as follows.
> 
> > Can we do this gradually and split things up a bit? Like, can this be done with each of the patches individually? That'd make the review easier, I think.
> 
> I initially submit this pull request as whole, for this modification affects depenency path of downstream mainboard crates. Commits submitted first modified project structure, it would result in failure of compilation for mainboard crates. The following commit fixes this problem by synchronizing structure in Cargo.toml dependency files. If this caused difficulty in review, I'd split this pull request into three, but first and second pull request won't build and pass CI checks.

I see, let's keep it all here in one PR then. We might squash when merging so that every commit still builds. :-)

> > Regarding the directory names: Could you explain the reason for renaming drivers to oreboot-drivers? Can't we keep it as it is, just `drivers`?
> 
> Cargo crate names are what it represents in Rust source code and surrounding ecosystem. If we call it `drivers`, this name would be too general for Rust ecosystem projects to reference it in source code (we already did this in oreboot project, `tock-registers` is a register representation in `tock` project). Name `oreboot-drivers` would be a better name for a cargo crate and source code as it represents drivers that are specific to oreboot project, where other projects can make use of if necessary.

I see, thanks! I didn't know that Cargo crates have no namespaces, so people make them up by themselves with dashes; TIL:
https://samsieber.tech/posts/2020/09/registry-structure-influence/

> Thank you! If applicable, please inform me and I'll split this pull request into smaller parts.

Do you have a use case for using the code here in other projects? Otherwise, I would suggest to keep the names as they are for now and rename them in a separate PR. It would be a simple change, right?

What I could imagine would be splitting up register blocks and drivers, because they drivers depend on the model, but the register block is always the same for everyone. Would that make sense?

luojia65:
> I see, let's keep it all here in one PR then. We might squash when merging so that every commit still builds. :-)

Thanks, I'll squash these commits.

> Do you have a use case for using the code here in other projects? Otherwise, I would suggest to keep the names as they are for now and rename them in a separate PR. It would be a simple change, right?

Hello! If there's a case to use oreboot code for other projects, I'd suggest it be vendor defined oreboot implementation. If commercial vendor regard it not appropriate to open source their platform but still want to use oreboot on private platform, they may avoid forking the whole oreboot project. Instead, they may depend on open source oreboot project modules, and use them in commercial implementations. By this way commercial consumers reuse oreboot modules (SoCs, drivers, etc.) while still providing all oreboot features on their platforms.

If this is not the main case, I'll rename folder names and/or crate names of these modules.

> What I could imagine would be splitting up register blocks and drivers, because they drivers depend on the model, but the register block is always the same for everyone. Would that make sense?

That's an awesome way where Rust's ecosystem benefits each other! :) This is on a use case whether projects other than bootloader would depend on this project or not. Other drivers may depend on oreboot drivers, while wrapping the abstractions inside and provide another abstraction. Although this possibility is rare, we make one abstraction of drivers, and the community can choose the best one they need to depend on.

orangecms:
> > Do you have a use case for using the code here in other projects? Otherwise, I would suggest to keep the names as they are for now and rename them in a separate PR. It would be a simple change, right?
> 
> Hello! If there's a case to use oreboot code for other projects, I'd suggest it be vendor defined oreboot implementation. If commercial vendor regard it not appropriate to open source their platform but still want to use oreboot on private platform, they may avoid forking the whole oreboot project. Instead, they may depend on open source oreboot project modules, and use them in commercial implementations. By this way commercial consumers reuse oreboot modules (SoCs, drivers, etc.) while still providing all oreboot features on their platforms.
> 
> If this is not the main case, I'll rename folder names and/or crate names of these modules.

The license is GPLv2. That mitigates the issue with closed source forks, and we do encourage working with upstream to keep everything streamlined. :-)
So reuse outside oreboot is not applicable.

luojia65:
> The license is GPLv2. That mitigates the issue with closed source forks, and we do encourage working with upstream to keep everything streamlined. :-)
> So reuse outside oreboot is not applicable.

Yes, the current way have this cons that it includes too much change. I will try this approach instead: the folder name would be 'driver', and the crate name would be 'oreboot -driver'.

rminnich:
If this doesn't break riscv boards then let's go for it.

luojia65:
Hello! I squashed all commits, and now the directory name are `arch` and `drivers` only.

orangecms (dismissed):
Looking good from my side :)

orangecms:
To pick this up:

@rjoleary asked if there could be a mechanism to define rules for the features in the arch create so that e.g. ARM and x86 features would be prevented from being accidentally mixed. @luojia65 said this would be possible and I suggest we do that later. With the current code, it would just as well be possible, technically, I think, so mid term we're winning some addition in terms of static analysis AFAIUI.

rminnich (approved):
I'm willing to try this and see how we like it. oreboot is not in any large scale use yet and we ought to be able to experiment. If this brings in the author of rustSBI as a contributor, all the better.

felixsinger (commented):
Please write a proper commit message.

luojia65:
Hi @felixsinger ! I derived this commit message from squashed commits. If this is not the correct way I'd change them, but I don't know if all reviews and CI checks should start again.

luojia65:
Hello! I updated commit message. No source code files are changed in this commit.

orangecms (approved):

PR #560, opened 2022-04-11
Remove all references to FSP code from oreboot, including submodules …

aquadara:
…and docs


rminnich (changes requested):
oops! This is a bit too much :)

let's do the coffeelake, tigerlake, and upxtreme boards as one commit to start?

rminnich (approved):

rminnich (approved):

PR #566, opened 2022-04-19
Makefiles: tweak them for the new layout

rminnich:
The location of target has changed; for all mainboards, target
is at $(TOP)/oreboot.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,9 +142,4 @@ checkandbuildall: ciprepare clippy checkformat test mainboards
 	echo "Done CI!"
 
 clean:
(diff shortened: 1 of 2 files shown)


luojia65 (approved):
This change LGTM.

BTW, I'll note here as well that command `cargo clean` would do the same when it comes to cleaning target path. This would be useful for future xtask based building scripts.

orangecms (approved):

orangecms:
There is  more to do, including the test setup, but this is good for a start.

orangecms:
There is  more to do, including the test setup, but this is good for a start.

PR #571, opened 2022-05-20
Nukesifive

rminnich:


rminnich:
thinking about dhendrix comments:

- we need soc/sifive for the qemu target
- hifive did in fact work

I think I'll just try to get it back.

PR #582, opened 2022-06-11
Fix qemu riscv dts

tyalie:
The oreboot binary was large enough to not fit into the flash info area as defined by the qemu-riscv `fixed-dtfs.dts` if the binary contained debug information. Only the release build did fit. As we have enough memory over, we can increase the flash and even create a bootable image for debug.


orangecms (approved):
Oh whew, how did that get so huge? Looks like you found something that could use revisiting. 😅 Increasing the size sounds good to me!

tyalie:
Perfect. Should we do the merge then ^^

orangecms:
> Perfect. Should we do the merge then? ^^

Noone objected, sooo... 🥳

PR #583, opened 2022-06-16
document the boot flow

orangecms:


rminnich (commented):
I still wish the u-root logo were right there under the linuxboot logo.

Documentation/boot-flow.md
```diff
@@ -0,0 +1,56 @@
...
+boot loader environment. In oreboot, we implement SBI via [RustSBI](
+https://github.com/rustsbi/rustsbi), a Rust crate offering all the necessary
+functions we need as a library.
+
+### Stage 3: Payload (LinuxBoot)
+
+In this final stage, we are presented with a Linux environment. To have all the
+familiar commands, such as `ls`, `cat`, etc, as well as boot loaders, we choose
+[u-root](https://u-root.org) to build our initramfs. However, you may as well
+embed [cpud](https://github.com/u-root/cpu) directly, or your own custom app.
```
rminnich, line 54: do we have a network setup yet or can we say something about network? These boards are pretty useless without it.

rminnich (commented):
Why the two pictures?

orangecms:
> I still wish the u-root logo were right there under the linuxboot logo.

Let me do that. I didn't want to clutter the image too much, but I'll try.

Documentation/boot-flow.md
orangecms: Well, it "depends on the board". Some have Wi-Fi, some have ethernet, some may be extended, e.g., with a SPI ethernet or SDIO Wi-Fi module.

Could you maybe elaborate on how you picture to "have a network setup"? I have many different network setups in the above sense of "it depends".

SamYaple (dismissed):
as a newcomer, this has greatly improved my understanding of the general scope, flow, and terminology used in the project. The diagrams are always appreciated.

Documentation/boot-flow.md
```diff
@@ -0,0 +1,71 @@
...
+## Stages
+
+Due to the nature of platforms, firmware needs to start small. From stage to
+stage, we set up an environment such that an operating system may be booted or
+any other payload executed.
+With oreboot, we focus on embedding [LinuxBoot](https://linuxboot.org).
+
+![oreboot flash and flow](img/firmware-matryoshka.png)
+
+### Stage 1: SRAM/XIP
```
rminnich, line 29: The only issue is that on some systems the bootblock has to be distinct from this first stage, since not even the processor mode is compatible. I would still argue that we need a Stage 0: here, which can be optional?

rminnich (commented):
Still trying to understand why the .svg and .png differ, although I guess .png is .svg + some stuff?

orangecms:
> Still trying to understand why the .svg and .png differ, although I guess .png is .svg + some stuff?

They are the same, and the .drawio file is the common source.
It's just that the .png is a static artifact, whereas the .svg retains the references to the source images that are included (the logos) and depending on the renderer, the output may differ. Hence the .png in the .md to show the final image.

rminnich (approved):

PR #555, opened 2022-03-30
Organize oreboot project into cargo workspace

luojia65:
This pull request is a successor to https://github.com/oreboot/oreboot/pull/553, it organizes the Oreboot project into one cargo workspace.

After this pull request, you may:

- run `cargo fmt` to reformat, `cargo clippy` to clippy check all packages under oreboot
- run `cargo clean` to clean the build cache shared by all oreboot packages
- discover a significant compile speed boost if you build all boards

The build framework is still based on existing Makefile system, but this reorganization would allow us to write a xtask based build framework in the future, onto project that is arranged under a cargo workspace.

## Checklist

- [x] A proper commit message

## Details to this change

Previously, oreboot consists of individual cargo packages. This would comply with C code, but it can't share incremental build caches. In this commit we will organize it into one workspace for cargo, in this case cargo would use the same target path so compilation caches would be shared.

To orgainze workspace, we rearranged soc and cpu modules, as is done in driver module. This rearrangement does not include any change to working code except only for changes to dependency path.

This commit includes changes to build.rs file, it's mostly used in FSP build and link script discovery. Changes to them will allow to detect the 3rdparty directory and search for link.ld in every path of organized workspace, so we may build them wherever we are at in the project directory.

In Cargo `[profile.release]` is defined in workspace configuration, I remove them from each individual project or Cargo would warn us that this profile configuration is ignored as already been configurated by workspace.

The commit also includes multiple clippy fixes. The project after this commit would still need makefiles to help building, but this organization to workspace would furtherly allow us to develop a xtask based building framework onto oreboot project.

PR #585, opened 2022-06-19
Introduce the xtask building system

luojia65:
This pull request is a successor to #555, it introduces the `xtask` building system into Oreboot project. The xtask system provides an easy way to build, flash and debug Oreboot with more host platforms (like Windows). The Makefile build system is still kept for compatibility.

To use `xtask` build system, you may execute the following commands:

* cargo make
* cargo flash
* cargo gdb
* cargo asm

... to build image, flash, debug or view assembly code on Oreboot platforms.

On each command, we change directory in `src/mainboard/<vendor>/<board>` or use `--mainboard` to decide which board we need to build to, `--variant` to choose the type of target board, and use `--release` to build under release profile. Parameter `--verbose` would print verbose output message to allow us to debug xtask itself.

The xtask system currently supports sunxi/nezha platform which has not upstreamed yet. Use command `cargo run`, we run xtask to view a list of commands we may use.

Xtask system may also to expanded to run emulators within command line, this feature is reserved in current commit to allow future implementation on different emulation mainboards.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,9 @@
+[alias]
+xtask = "run --package xtask --"
+make = "xtask make"
+asm = "xtask asm"
+flash = "xtask flash"
+# debug = "xtask debug"
+gdb = "xtask gdb"
+
+# todo: xtask qemu for emulation
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,10 @@ members = [
     "src/mainboard/sunxi/*",
 
     "tools/layoutflash",
+
+    "xtask",
 ]
+default-members = ["xtask"]
 
 [profile.release]
 opt-level = 'z'  # Optimize for size.
--- a/xtask/Cargo.toml
+++ b/xtask/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "xtask"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+clap = { version = "3.1.18", features = ["derive"] }
+clap-verbosity-flag = "1.0"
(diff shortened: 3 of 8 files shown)


.cargo/config.toml
```diff
@@ -0,0 +1,9 @@
+[alias]
+xtask = "run --package xtask --"
+make = "xtask make"
+asm = "xtask asm"
```
orangecms, line 4: Let's call this target `objdump` so it's clear what it does, WDYT?

luojia65: Cargo objdump is surely better than cargo asm, but this overlaps a command of widely used `cargo-binutils` tool at: https://github.com/rust-embedded/cargo-binutils/tree/master/src/bin .

orangecms: Oh interesting, I didn't know - good point :-)

orangecms: What is the difference between the objdumps? I can't see it yet.
Like, all toolchains have an objdump, be it GCC, LLVM, or the wrapper in Cargo.

orangecms: So this here just invokes `llvm-objdump` etc:
https://github.com/rust-embedded/cargo-binutils/blob/83cc6c14d40c6db463e10be3e36e1a43846d58cd/src/tool.rs#L43

Now `objdump`, for a given binary with metadata (e.g., ELF), it gives you the disassembled parts, whereas `asm`, on the other hand, may have people think of doing the forward direction, assembly.

orangecms: Ah, are you saying we cannot use `cargo objdump` as an alias because it already exists?

How do you feel about `disasm`?

orangecms: Let's take it as it is for now :)

orangecms (approved):
Let's take this as it is for now and work out more details over time. Thank you! 🥳

PR #586, opened 2022-06-25
sunxi/nezha: decompress LinuxBoot by default

orangecms:

### DIFF
--- a/src/mainboard/sunxi/nezha/src/main.rs
+++ b/src/mainboard/sunxi/nezha/src/main.rs
@@ -1,10 +1,10 @@
 #![no_std]
 #![no_main]
-#![feature(default_alloc_error_handler)]
+#![feature(default_alloc_error_handler, ptr_metadata, allocator_api)]
 
-use core::arch::global_asm;
 use core::fmt::Write;
 use core::panic::PanicInfo;
+use core::{arch::global_asm, ptr::read_volatile, slice};
 use oreboot_arch::riscv64 as arch;
 use oreboot_drivers::{
     uart::sunxi::{Sunxi, UART0},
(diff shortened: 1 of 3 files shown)


rminnich (commented):
so, random thought. The driver model was designed for things like this. If you were to have the input be provided by a Device, e.g. a memory slice, and the output be a device, similarly a memory slice, there'd be pretty interesting stuff we could do, including having input come from a block Device or serial line.

For later. Just at thought.

orangecms:
> so, random thought. The driver model was designed for things like this. If you were to have the input be provided by a Device, e.g. a memory slice, and the output be a device, similarly a memory slice, there'd be pretty interesting stuff we could do, including having input come from a block Device or serial line.
> 
> For later. Just at thought.

Yea, that's what I meant by what I wrote in Slack. :-)

src/mainboard/sunxi/nezha/src/main.rs
```diff
@@ -37,49 +101,27 @@ pub extern "C" fn _start() -> ! {
...
     } else {
-        let entry = mem + payload_offset;
-        let payload: payload::Payload = payload::Payload {
+        // TODO; This payload structure should be loaded from boot medium rather
+        // than hardcoded.
+        let entry = MEM + PAYLOAD_OFFSET;
+        let segs = &[payload::Segment {
+            typ: payload::stype::PAYLOAD_SEGMENT_ENTRY,
+            base: CACHED_MEM,
+            data: &mut SectionReader::new(&Memory {}, entry, PAYLOAD_SIZE),
```
rminnich, line 122: See this line? In principle, if the decompressor were a Device, the data here could be a &mut DecompressReader::new(device: memory section reader, or disk, or serial line, depending on jumpers);

There's some power there it would be nice to use. But it's more work for sure ... 

anyway ... I think this should go in as is, because it is the difference between usable and unusable kernel payload?

orangecms:
superseded by #590

PR #590, opened 2022-07-01
full xtask build for sunxi/nezha

orangecms:
Now a D1 board can be built like this:

```sh
 cargo flash --variant lichee --memory nor \
   --payload linux/arch/riscv/boot/Image \
   --dtb linux/arch/riscv/boot/dts/allwinner/sun20i-d1-lichee-rv.dtb
```

This will build two stages, bt0 (running in SRAM) and main (in DRAM), wrap the eGON header around bt0 as required by the SoC, and add the compressed Linux kernel as a payload (use the raw Image, not a gz!) plus its dtb.

Note that for the time being the Linux kernel needs to come from a fork with hardcoded values for DRAM size. Upstream doesn't have all the D1 patches yet.

See here for building a LinuxBoot image, ignoring the parts on `xfel`; those are obsolete with this PR:
https://gist.github.com/orangecms/df0d3c15f9e32b5c708c021c4be8857b#5180-rc


rminnich (approved):

orangecms:
getting this
```
error: language item required, but not found: `eh_personality`
  |
  = note: this can occur when a binary crate with `#![no_std]` is compiled for a target where `eh_personality` is defined in the standard library
  = help: you may be able to compile for a target that doesn't need `eh_personality`, specify a target with `--target` or in `.cargo/config`

warning: `qemu-q35` (bin "qemu-q35") generated 1 warning (1 duplicate)
error: could not compile `qemu-q35` due to previous error; 1 warning emitted
```

need to check https://docs.rust-embedded.org/embedonomicon/smallest-no-std.html#eh_personality

rminnich (approved):
I'm ready if you are.

PR #599, opened 2022-08-23
Drop all mainboards save emulation and sunxi/nezha

rminnich:
We are moving to the embedded hal and other common crates.
This entails dropping our internal driver crates; we've found
that a driver model is not that crucial.

As a first step, remove mainboards that still use the old model.

This being git, they are always there, and we can bring them back
when we are ready.


orangecms (approved):
Alright, let's go :boom: all!

PR #601, opened 2022-08-30
Move old style mainboard to mainboard.broken so we can remove old lib…

rminnich:
…raries

That way our CI will continue to work as we make these changes.


orangecms (dismissed):

orangecms:
Hum, Clippy is unhappy.

Let's move those to `src.broken/mainboards` - WDYT?

luojia65:
LGTM

PR #621, opened 2022-11-01
new ci task: format/lint all toml, json, markdown

bwerthmann:
* Add CI action for dprint formatter
  * <https://dprint.dev/>
* Add an initial dprint config
  * `cargo install dprint`
  * `dprint init`
  * exclude target directory
* Create a new CI job for better concurrency
* The only "net new" items I've commented on are in 49e707ae59b190333bc0c5c6c62a4769de861db8. All other changes should be no-op whitespace changes.
* Here's the shell one-liner I used to generate these commits:

```bash
(set -x; dprint output-file-paths |sort |while read fn; do gfn=$(sed 's/\/home\/ben\/ofw\/oreboot-cicd\///' <<<"$fn"); dprint fmt --diff --verbose -- "$gfn" ; git add "$gfn" && git commit -s -m "$gfn : consistent format from dprint" ;done)
```

## Effective Config

The effective config can be seen with `dprint output-resolved-config`. Changes can be made if we prefer something else.

```json
{
  "json": {
    "array.preferSingleLine": false,
    "commentLine.forceSpaceAfterSlashes": true,
    "ignoreNodeCommentText": "dprint-ignore",
    "indentWidth": 2,
    "lineWidth": 120,
    "newLineKind": "lf",
    "object.preferSingleLine": false,
    "useTabs": false
  },
  "markdown": {
    "emphasisKind": "underscores",
    "ignoreDirective": "dprint-ignore",
    "ignoreEndDirective": "dprint-ignore-end",
    "ignoreFileDirective": "dprint-ignore-file",
    "ignoreStartDirective": "dprint-ignore-start",
    "lineWidth": 80,
    "newLineKind": "lf",
    "strongKind": "asterisks",
    "textWrap": "maintain"
  },
  "toml": {
    "cargoApplyConventions": true,
    "commentForceLeadingSpace": true,
    "indentWidth": 2,
    "lineWidth": 120,
    "newLineKind": "lf",
    "useTabs": false
  }
}
```

## Next Up

`dprint` has plugins for:

- [ ] #623 [Rustfmt](https://dprint.dev/plugins/rustfmt)

It can also [Exec](https://dprint.dev/plugins/exec) format tools should we need to do so. 

This PR is large enough as-is.


bwerthmann:
Here's the shell one-liner I used to generate these commits:

```bash
(set -x; dprint output-file-paths |sort |while read fn; do gfn=$(sed 's/\/home\/ben\/ofw\/oreboot-cicd\///' <<<"$fn"); dprint fmt --diff --verbose -- "$gfn" ; git add "$gfn" && git commit -s -m "$gfn : consistent format from dprint" ;done)
```

.github/workflows/build.yml
```diff
@@ -12,6 +12,14 @@ env:
...
 
 jobs:
+  QuickFormat:
+    runs-on: ubuntu-latest
+    steps:
+      - name: git-checkout-action
+        uses: actions/checkout@v3
+      - name: dprint-check-action
+        uses: dprint/check@v2.1
+
```
bwerthmann, line 22: New CI Job for better concurrency:

https://github.com/oreboot/oreboot/actions/runs/3371817811/jobs/5594518414#step:3:14

dprint.json
```diff
@@ -0,0 +1,18 @@
...
+  "excludes": [
+    "**/*-lock.json",
+    "target/"
+  ],
+  "plugins": [
+    "https://plugins.dprint.dev/json-0.16.0.wasm",
+    "https://plugins.dprint.dev/markdown-0.14.1.wasm",
+    "https://plugins.dprint.dev/toml-0.5.4.wasm"
+  ]
+}
```
bwerthmann, line 18: dprint config file

bwerthmann:
@orangecms @rminnich PTAL

.github/workflows/build.yml
rminnich, line 22: I've seen recommendations that tools like dprint are fine, but using them in CI is not as fine, b/c the output of these tools vary widely, even for a given version, on different distros.

how about you do the dprint, and commit what it changes, and it can even be a top level Makefile target or other tool's target, but it's not a check in CI?

bwerthmann: > ...it can even be a top level Makefile target or other tool's target

I agree we should do this. It was easier to initially get this up and running in CI than digging into the `make + xtask bailiwick`.

A task for myself:

- [ ] Have a top level make task for  the codebase, for developer ease of use. Ideally, the `fmt`ers should be run as a pre-push hook.

> I've seen recommendations that tools like dprint are fine, but using them in CI is not as fine, b/c the output of these tools vary widely, even for a given version, on different distros.

In recent history I'd agree with you on this point. However, in this case we have a stable `fmt`ing toolchain. 

* `cargo install dprint`. `dprint` just has the search mechanics and CLI frontend.
* dprint plugins refs are platfrom independent wasm payloads and are version locked in git. They are also rust programs.
```json
  "plugins": [
    "https://plugins.dprint.dev/json-0.16.0.wasm",
    "https://plugins.dprint.dev/markdown-0.14.1.wasm",
    "https://plugins.dprint.dev/toml-0.5.4.wasm"
  ]
```
* Historically, I agree with you, does this information sway you one way or another?

> how about you do the dprint, and commit what it changes
> (...) but it's not a check in CI?

* IMHO it's easier for everyone for us to be opinionated about the "`fmt`-ing" we accept. Then we don't have 100s or 1000s of whitespace changes based on whomever edited the files last. This makes life (rebase/history/review/merges) easier because change sets have exactly one representation. An exception is for any generated artifacts (lock files for example). 


**...As a middle ground, we can configure the "`fmt`-ing" CI task to be advisory, like a compiler warning.**

bwerthmann: ...I was inspired to do this based on:

https://github.com/oreboot/oreboot/blob/57f2d72e1ff67c54ffd10fc744d69fbd5d21386c/README.md?plain=1#L216-L217

bwerthmann:
The effective config can be seen with `dprint output-resolved-config`. Changes can be made if we prefer something else.

```json
{
  "json": {
    "array.preferSingleLine": false,
    "commentLine.forceSpaceAfterSlashes": true,
    "ignoreNodeCommentText": "dprint-ignore",
    "indentWidth": 2,
    "lineWidth": 120,
    "newLineKind": "lf",
    "object.preferSingleLine": false,
    "useTabs": false
  },
  "markdown": {
    "emphasisKind": "underscores",
    "ignoreDirective": "dprint-ignore",
    "ignoreEndDirective": "dprint-ignore-end",
    "ignoreFileDirective": "dprint-ignore-file",
    "ignoreStartDirective": "dprint-ignore-start",
    "lineWidth": 80,
    "newLineKind": "lf",
    "strongKind": "asterisks",
    "textWrap": "maintain"
  },
  "toml": {
    "cargoApplyConventions": true,
    "commentForceLeadingSpace": true,
    "indentWidth": 2,
    "lineWidth": 120,
    "newLineKind": "lf",
    "useTabs": false
  }
}
```

rminnich (approved):

PR #636, opened 2022-11-09
Initial SD Card Support 

bwerthmann:


xtask/src/sunxi/nezha.rs
```diff
@@ -336,6 +344,10 @@ fn xtask_burn_d1_flash_bt0(xfel: &str, env: &Env) {
     let mut command = Command::new(xfel);
     command.current_dir(dist_dir(env, DEFAULT_TARGET));
     match env.memory {
+        Some(Memory::MMC) => {
```
rminnich, line 347: I would prefer that writing the SD from oreboot never be implemented ...

PR #638, opened 2022-11-11
tools/layoutflash: bring it back

rminnich:
This is working as it did before, but the code is much simpler.


tools/layoutflash/src/main.rs
```diff
@@ -1,46 +1,53 @@
...
+// offset 0 items will be placed first!
+// So, new rules:
+// This method assumes that the areas are specified in order for a reason.
+// In most cases, the offset is 0, meaning "use the previous area end+1
+// as our offset". In rare cases, an offset is specified; for that, ensure
+// there is no overlap with previous Areas. All Areas after this Area will
+// be placed after it. That way, only a limited number of offsets need
+// to be specified, possibly even 0, and the order in the ROM image will be the order
+// specified in the DTS.
+fn layout_flash(path: &Path, areas: &mut Vec<Area>) -> io::Result<()> {
```
rmsyn, line 36: ``
```suggestion
fn layout_flash(path: &Path, areas: &mut [Area]) -> io::Result<()> {
```

minor style, since no `Vec` specific functionality is used

orangecms (approved):
let's just get this in

tools/layoutflash/src/main.rs
luojia65: This will make the software more Rusty, I'd suggest we apply this commit

Issue #551, opened 2022-02-25
Unstable features and macros tracking issues


orangecms:
Same as they are doing in Rust for Linux: https://github.com/Rust-for-Linux/linux/issues/2

Browse the [Rust unstable book](https://doc.rust-lang.org/stable/unstable-book) to find the references.

Here's a list of unstable things we're using (TODO: complete the list!)

## Features

- [build-std](https://doc.rust-lang.org/cargo/reference/unstable.html#build-std)

## Macros, language features

- [asm_const](https://doc.rust-lang.org/stable/unstable-book/language-features/asm-const.html)

bwerthmann:
I spiked on this for just the nezha as far as I could get with `make RUSTUP_TOOLCHAIN=stable`. See branch https://github.com/bwerthmann/oreboot/tree/551-unstable-features-and-macros-tracking-issues for patches. I only compile tested this code.

orangecms:
Just a note here:

- `asm_const` and naked functions are stable these says
- no memory allocation besides stack is necessary for us, really

PR #656, opened 2022-12-16
build Nezha with 1.66 (stable) toolchain

bwerthmann:
Experimental. This is as far as I made it with `make RUSTUP_TOOLCHAIN=1.66`.

Relates to #551


orangecms:
We have moved on to newer and newer Rust toolchains.
There are a few things left until we can move to stable.
And we can already move on to Rust edition 2024 with some adjustments.
Closing this here now.

PR #659, opened 2022-12-31
sipeed/m1s-dock: bootstrap board

orangecms:


rminnich (commented):
it's been almost 4 years since we started this project, and I'm starting to feel like the code is converging to a nice new place. This is nice.

rminnich (commented):
if you need all of main to be unsafe?

orangecms:
> if you need all of main to be unsafe?

no longer with a fixup in the PAC crate; PR pending: https://github.com/bouffalolab/bl808-pac/pull/5

orangecms:
and with today's changes, both cores are running now (note: heterogeneous - 32bit and 64bit!) :)

orangecms:
Checkpoint: PSRAM works just fine right now.

orangecms:
I'll factor out memtest to lib/, and squash things down here
otherwise it is good to go

mind that the PSRAM init can apparently be simplified, see
https://github.com/rustsbi/bouffalo-hal/blob/5d3dd7a56cf3c88c999cc901a919aed6e5bb6782/bouffalo-hal/src/psram.rs#L23

PR #660, opened 2023-01-11
Bump version of heapless

rjoleary:
This resolves the dependabot warnings.

Some Cargo.lock files were for broken mainboards, thus I was unable to generate them easily. I decided to simply delete those lock files.


orangecms (approved):

orangecms:
The lock file deletions are just fine - we got rid of the multiple ones with the switch to a Cargo workspace, and now there is only one left on the root.

Thanks a lot! :)

PR #665, opened 2023-01-24
src/arch: remove

orangecms:
This is no longer needed. We use the riscv crate.
If someone wants to bring in PPC64, look for a suitable crate, e.g., https://docs.rs/core_arch/0.1.5/core_arch/powerpc64/index.html


rmsyn:
Certain architectures in other projects use a `start.S`, `crt0.S` for initial bring-up and runtime setup.

There are some `-rt` crates out there ([riscv-rt](https://github.com/rust-embedded/riscv-rt), [cortex-m-rt](https://github.com/rust-embedded/cortex-m/tree/master/cortex-m-rt)), but not for every architecture (`aarch64` has no `cortex-a-rt` in [aarch64-cpu](https://github.com/rust-embedded/aarch64-cpu)).

Where should the startup code (assembly, or rustified startup) live while startup/runtime crates are being developed?

rminnich:
after watching this for a few months: 
let's not delete src/arch, we can leave it there, and have it reference other crates as needed. That will allow us to have local customizations, e.g., the startup code for x86 in oreboot was the best I've seen, so we keep that local for now. 

So the work here can be to remove the parts we know we don't want, and leave in the parts we need atm, and evolve.

sound ok?

orangecms:
in favor of #691

PR #667, opened 2023-01-24
sunxi/nezha: drop allocation

orangecms:
Simply `print!` instead of allocating `Vec`s, which is not necessary. This simplifies the code and enables adding the lib/log crate, which implies allocation itself and otherwise clashes with the global allocator used in hart_csr_utils.


orangecms:
Note: this change saves some kilobytes in build size and gets us rid of a dependency.

bwerthmann (approved):
Nice cleanup :+1:

dancrossnyc (commented):
LGTM

PR #668, opened 2023-01-24
sunxi/nezha: switch to lib/log (part1)

orangecms:
In this first step, replace the `uart` implementation in `soc/d1`.

Next see if we can simplify `sunxi/nezha/main` and get rid of the wrapping.

This requires #667 first.


orangecms:
cleanup to follow

rminnich (dismissed):

rminnich (approved):
nice evolution.

PR #606, opened 2022-09-07
oreboot visionfive1 startover

orangecms:
- create new mainboard starfive/visionfive1
- starfive/visionfive1: bootstrap bt0 stage

PR #686, opened 2023-02-27
bootstrap mainbords/starfive/visionfive2

orangecms:
Note:

This is currently based on the VF1 / JH7100 rework PR. The boards and SoCs are very similar wrt what needs to be done in oreboot.


src/mainboard/starfive/visionfive1/bt0/src/main.rs
```diff
@@ -0,0 +1,247 @@
...
+#[link_section = ".bss.uninit"]
+static mut BT0_STACK: [u8; STACK_SIZE] = [0; STACK_SIZE];
+
+/// Set up stack and jump to executable code.
+///
+/// # Safety
+///
+/// Naked function.
+#[naked]
+#[export_name = "start"]
```
rmsyn, line 66:
```suggestion
#[no_mangle]
```

src/mainboard/starfive/visionfive1/bt0/src/main.rs
```diff
@@ -0,0 +1,247 @@
...
+    /* ONCE.call_once(|| */
+    /*
+    unsafe {
+        static mut SERIAL: Option<JH71XXSerial> = None;
+        SERIAL.replace(s);
+        log::init(SERIAL.as_mut().unwrap());
+    }*/ //);
+}
+
+fn main() {
```
rmsyn, line 143:
```suggestion
#[no_mangle]
fn main() {
```

src/mainboard/starfive/visionfive2/bt0/src/init.rs
```diff
@@ -0,0 +1,548 @@
...
+    */
+
+    init_coreclk();
+
+    // These take like 16 cycles to actually propagate. We can't go sending
+    // stuff before they come out of reset. So wait.
+    // TODO: Add a register to read the current reset states, or DDR Control
+    // device?
+    for _ in 0..=255 {
+        unsafe { asm!("nop") }
```
rmsyn, line 164:
```suggestion
        riscv::asm::nop()
```

src/mainboard/starfive/visionfive2/bt0/src/main.rs
```diff
@@ -0,0 +1,297 @@
...
+    //    intrinsics::transmute,
+    panic::PanicInfo,
+    ptr::{read_volatile, write_volatile},
+    // slice,
+};
+use riscv::register::mhartid;
+// use riscv::register::{marchid, mimpid, mvendorid};
+// use embedded_hal::serial::nb::Write;
+// use init::{clock_init, iopad_init, rstgen_init};
+use uart::JH71XXSerial;
```
rmsyn, line 20: Might be good to use the [uart_16550](https://docs.rs/uart_16550/latest/uart_16550/#with-memory-mapped-serial-port) crate here. Would work for both `visionfive1/2`, and other memory-mapped serial boards.

orangecms: Yea that one has its own hardcoded values though, and atomic instructions do not work on the JH7100 outside DRAM, i.e., this thing may go nuts in SRAM. That crate is rather for OSes than firmware.
See also:
https://github.com/rust-osdev/uart_16550/blob/d401ea103641cc06182c404840a5ba2896eebe72/src/mmio.rs#L52

We may just have our own internal crate instead, and I think we will. I am looking at multiple SoCs to figure this out. :-)

rmsyn: Makes sense, didn't know if there was enough crossover to be usable in firmware.

Also didn't know about atomics not working in SRAM...

> We may just have our own internal crate instead

That would be great!

src/mainboard/starfive/visionfive2/bt0/src/init.rs
```diff
@@ -0,0 +1,548 @@
...
+pub const SYSCON_IOPAD_CTRL100: u32 = SYSCON_IOPAD_CTRL_BASE + 0x190;
+pub const SYSCON_IOPAD_CTRL101: u32 = SYSCON_IOPAD_CTRL_BASE + 0x194;
+pub const SYSCON_IOPAD_CTRL102: u32 = SYSCON_IOPAD_CTRL_BASE + 0x198;
+pub const SYSCON_IOPAD_CTRL103: u32 = SYSCON_IOPAD_CTRL_BASE + 0x19C;
+pub const SYSCON_IOPAD_CTRL104: u32 = SYSCON_IOPAD_CTRL_BASE + 0x1A0;
+
+pub fn syscon_func_0(v: u32) {
+    // NOTE: for whatever reason, it appears that writing only works after
+    // reading i.e., if you remove the `read_32`, it breaks the code
+    // let's hope the compiler does not remove it in optimization
```
rmsyn, line 235: Maybe it has something to do with cached writes or similar? Would a `fence`, `write_volatile`, or other memory barrier allow for removing the `read_32`?

Not sure how much difference it makes, but might help with readability...

rmsyn, line 235: Maybe it has something to do with cached writes or similar? Would a `fence` or other memory barrier allow for removing the `read_32`?

Not sure how much difference it makes, but might help with readability...

orangecms: Note: This is copied from the VF1 implementation and may be entirely different for the JH7110.

We are running on a single thread without interrupts here. Not sure what a memory barrier would be for. I had to read the register once, from what I found back then. May also be something else, such as alignment.

rmsyn: > Not sure what a memory barrier would be for.

Right, I meant memory ordering instructions, my bad. Would `fence.i` help? It says it makes writes available to memory fetching instructions on the same hart. Not sure it would have any effect on write instructions.

orangecms: Sorry, I don't understand what this would mean here....
We just have to read that register. I do not see how that is related with anything like harts or synchronitation. It's about the behavior of the peripheral. We are on a single hart with no interrupts here.

rmsyn: > Sorry, I don't understand what this would mean here....

Yeah, I was ultimately just bike-shedding here. Probably not worth it to figure out why the hardware requires the read before write. It just bugged me, because it's weird.

src/mainboard/starfive/visionfive2/bt0/src/init.rs
```diff
@@ -0,0 +1,677 @@
...
+    unsafe {
+        write_volatile(reg as *mut u32, val);
+    }
+}
+
+fn read_32(reg: u32) -> u32 {
+    unsafe { read_volatile(reg as *mut u32) }
+}
+
+pub const CLKGEN_BASE: u32 = 0x1180_0000;
```
rmsyn, line 143:
```suggestion
pub const CLKGEN_BASE: u32 = 0x1302_0000;
```

https://doc-en.rvspace.org/JH7110/TRM/JH7110_TRM/system_memory_map.html
https://github.com/starfive-tech/u-boot/blob/1539c1fb5a498fd7d20c2cffc84c187d703c8b66/arch/riscv/dts/jh7110.dtsi#L268

src/mainboard/starfive/visionfive2/bt0/src/init.rs
```diff
@@ -0,0 +1,677 @@
...
+        if !v != 0x1 {
+            break;
+        }
+    }
+}
+
+pub fn clear_rstgen_rstn_gmac_ahb_() {
+    let mut v = read_32(RSTGEN_SOFT_ASSERT1);
+    v &= !(0x1 << 28);
+    v |= 0 << 28;
```
rmsyn, line 609: Isn't this a no-op?

orangecms: Yes, and you need not care about this... this whole chunk here is copied from the previous SoC and just here for reference... may not be applicable at all. This is all a draft, and what I am currently on is figuring out issues with print macros...

src/mainboard/starfive/visionfive2/bt0/src/init.rs
orangecms: I know. This is copied from the previous SoC and only here for reference. We are not at the point where we deal with clocks. The current work is around issues with print macros.

rmsyn: Ok, just following along, and noticed this bit. If it's more helpful, I can hold off on review until a later point.

src/mainboard/starfive/visionfive2/bt0/src/init.rs
rmsyn: If it's a no-op, and serves no purpose, wouldn't it be better to just remove it?

> This is all a draft, and what I am currently on is figuring out issues with print macros...

Understood, if you want debugging/testing help, let me know.

orangecms: See the latest commits I pushed.
For reference, this happens:
https://twitter.com/OrangeCMS/status/1632208849254981633

src/mainboard/starfive/visionfive2/bt0/src/init.rs
orangecms: We went through this a bit in today's live stream. I have defined a new constant `SYS_CRG_BASE`, aligning with the docs. DDR init is the next step now; printing works fine after fixing the SRAM address in the linker script.

src/mainboard/starfive/visionfive2/bt0/src/dram.rs
```diff
@@ -0,0 +1,100 @@
...
+
+struct MemCfg {
+    offset: u32,
+    mask: u32,
+    value: u32,
+    flags: u32,
+}
+
+impl MemCfg {
+    fn new(offset: u32, mask: u32, value: u32, flags: u32) -> Self {
```
rmsyn, line 19:
```suggestion
    pub const fn new(offset: u32, mask: u32, value: u32, flags: u32) -> Self {
```

src/mainboard/starfive/visionfive2/bt0/src/dram.rs
```diff
@@ -0,0 +1,100 @@
...
+///         mask: 0x0,
+///         value: 0x40001030,
+///         flags: (OFFSET_SEL | F_SET | REG2G),
+///     },
+/// ];
+/// ```
+macro_rules! mem_cfg_arr {
+    ($({ $o: expr, $m: expr, $v: expr, $f: expr }),* $(,)?) => {
+        [
+            $(MemCfg { offset: $o, mask: $m, value: $v, flags: $f }),*
```
rmsyn, line 67:
```suggestion
            $(MemCfg::new($o, $m, $v, $f as u32)),*
```

With the `MemCfg::new` being a `const fn` you could probably get rid of the macro. You would just have to cast the last argument to `u32` on each call below.

src/mainboard/starfive/visionfive2/bt0/src/dram.rs
```diff
@@ -0,0 +1,100 @@
+// NOTE: Those _could_ be an enum, but this makes it easily compatible with the
```
rmsyn, line 1:
```suggestion
use core::ptr::{read_volatile, write_volatile};

// NOTE: Those _could_ be an enum, but this makes it easily compatible with the
```

src/mainboard/starfive/visionfive2/bt0/src/main.rs
rmsyn: Just opened a pull request to allow custom UART init in [uart_16550](https://github.com/rust-osdev/uart_16550/pull/25).

If we can solve the issue with atomics being wonky in SRAM, maybe we can use their lib in `oreboot`?

Could also try replacing their atomic use with a `SpinLock`, or a wrapper that conditionally enables memory barriers with a feature. That way we could disable for the case of the bootloader running on a single core, single thread.

orangecms: Do you have a VF2 to experiment with? Would be nice if you could try this out. I'm currently translating the DRAM init, which takes quite a while...

rmsyn: > Do you have a VF2 to experiment with?

Yes, I'll try flashing what you have now. Is the current stuff you have bootable?

Also, I was going to start working on a QEMU setup for testing. Does the current emulation board work for VF2?

orangecms: The current code is neither for flashing nor QEMU.

You can go back some commits to have something that can be run from UART. Note that there is no clock/PLL setup yet, so we just pick up what the mask ROM did.
See also the notes in the Makefile.

orangecms: The cureent QEMU board is linked at an address that differs from the JH7110's SRAM. And it uses the ns16550 crate, which is unlikely to work. Feel free to experiment though.

Remember to wrap the binary in the necessary header.
I recommend the UART loader instead of flashing for convenience.
See also the recordings for reference:
https://youtube.com/playlist?list=PLenOHeTI_A9MJlYIOAVC0JDpKKXX9mZgK

rmsyn: > The cureent QEMU board is linked at an address that differs from the JH7110's SRAM. And it uses the ns16550 crate, which is unlikely to work.

Yeah, I started looking at the QEMU board, and docs on the main QEMU site. I didn't know QEMU was so much more specific for ARM and RISCV boards (opposed to x86-64). Makes sense with ARM/RISCV being more embedded boards with wildly different peripherals and memory setup.

> I recommend the UART loader instead of flashing for convenience.

Awesome, I'll check out the linked video, and try to see how far I get.

src/mainboard/starfive/visionfive2/bt0/src/ddr_start.rs
```diff
@@ -2,7 +2,7 @@ use crate::ddrlib::{MemCfg, MemSet};
 use crate::init::{self, read32, write32};
 
 // TODO: define build time parameters (!)
-const VAL_X1: u32 = if cfg!(dram_size = "8G") || cfg!(dram_size = "4G") {
+const VAL_X1: u32 = if cfg!(dram_size = "8G") || cfg!(dram_size = "4G") || true {
```
rmsyn, line 5: Is this temporary? These will always evaluate to `true`.

orangecms: Yea. It's one commit hardcoding a 4GB config (what I have) for now. The cfg somehow didn't make it into the build.

The DRAM init is not working yet, missing some clocks maybe - I'll check today.

src/mainboard/starfive/visionfive2/bt0/src/dram.rs
orangecms: Note that this was intermediate code.

At this point, we have working DRAM init.

src/mainboard/starfive/visionfive2/bt0/src/ddr_start.rs
orangecms: I've rework this to a constant. If you can figure out why the `cfg!()` stuff doesn't work, please file a PR.

Eventually, the DRAM size would need to be read from an EEPROM.

orangecms:
This is now largely done. Note that #606  is the base and goes _first_.

rminnich (approved):

PR #688, opened 2023-03-08
arch: remove references to non-existent modules

orangecms:


rminnich:
why draft? Also, is there a crates.io crate we could use for arch/?

rminnich (approved):

PR #689, opened 2023-05-21
visionfive2 main.rs: use let Some(), not unwrap

rminnich:
Unwrap can be very dangerous in firmare; as code
evolves, we have found it can explode in unexpected ways.

Use let some instead.


orangecms:
This is a very interesting design cha(lle)nge.

Essentially, this is exchanging a panic for silently ignoring the error. In other words, we would be left without the serial initialized, but could still continue. That potentially has implications on using the `print` macros. Is it safe to use those?
As we are using the `if let Some()` pattern here, we should be fine:
https://github.com/oreboot/oreboot/blob/28d0eea9af3d0d3eeb26350d22ffe1491e89d82b/src/lib/log/src/lib.rs#L135

Which means that we would continue with the DRAM init and eventually call the next stage/payload, which may be able to initialize the serial / its print and logging again. :-)

orangecms (approved):

PR #696, opened 2023-06-09
move nezha compression library from mainboard to lib

rminnich:


src/mainboard/sunxi/nezha/main/Cargo.toml
```diff
@@ -17,8 +17,17 @@ vcell = "0.1.3"
...
-oreboot_compression = { path = "../compression" }
+
+[dependencies.oreboot_compression]
+path = "../../../../lib/compression"
+
+[dependencies.oreboot-soc]
+path = "../../../../soc"
+features = ["sunxi_d1"]
+
+[dependencies.riscv]
```
orangecms, line 28: this is already here further up - the dependencies are divided in 2 blocks: first external crates, then local (oreboot's own) crates

suggestion: let's keep the inline style; the multiline style seems a bit confusing, resembling different sections

rminnich: oh I thought that was your stuff

orangecms (approved):
not being too picky about the dependencies at this point - let's just get it in

PR #697, opened 2023-06-14
Libfdt

rminnich:


rminnich:
note this is not quite ready; tests are not done. But I pushed it to show the direction I think we can go.

bt0 and main.rs won't really be in the fdt, because I know of no way to write an fdt where data comes first and node description after. If I'm wrong be sure to tell me.

I figure bt0 and main get pasted together as we do now, and the fdt is concatenated to the end of main. Main will use the fdt crate, and search the fdt which starts on a 16-byte alignment after end() in main(). If that makes no sense lmk.

rminnich:
this needs many commits squashed. but it does what it ought: uses dts to drive how flash is laid out, and includes dtb in the image so it can be used to locate payloads.

rminnich:
what's with the CI? it builds locally for me. The error is not very clear.

rminnich:
builds a working image for visionfive2
passes CI

that was fun.

orangecms (approved):
Let's get it ~~on~~ in!

orangecms:
Squashed it all down. I'll file follow-ups refactoring some bits etc.

PR #698, opened 2023-06-18
sbi print macro usage fixup

orangecms:


rminnich (approved):

rminnich (dismissed):
I would kind of prefer the nezha fix be separate but your call.

orangecms:
Yea it's a separate commit, but but... I messed up here - adding more prints wasn't the intention. Sorry! Changed to draft, fixup to come. I may push a separate PR for the nezha.

orangecms:
okay so I've reduced this to only the print macro fixup

rminnich (approved):

PR #705, opened 2023-08-15
[qemu-riscv] Build fix

oddcoder:
This patch set fixes a collection of bugs in the build process of qemu-riscv target. Even thought the qemu-riscv target builds and to some extent works, that souldn't have done, and the only reason it worked previously is sheer luck. The initial boot starter assembly blob was never used. In addition to that, The load address for the binary was not correct and some unnecessary symbols were forced to be kept which gave some linker error once we start doing anything remotely interesting in the rom.

### DIFF
--- a/src/mainboard/emulation/qemu-riscv/link.ld
+++ b/src/mainboard/emulation/qemu-riscv/link.ld
@@ -26,7 +26,7 @@ ENTRY(_boot);
 
 SECTIONS
 {
-    . = 0x20000000;
+    . = 0x80000000;
     .bootblock :
     {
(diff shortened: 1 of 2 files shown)


orangecms:
sorry - this and other PRs just ran into this dprint thing (our formatter, CI uses its latest version) https://github.com/dprint/dprint/issues/731 - I'm filing a PR to upgrade thing

orangecms:
see https://github.com/oreboot/oreboot/pull/706 for the CI fix

orangecms (approved):

PR #708, opened 2023-08-19
Run compiled binary on Lichee

m0veax:
Hey,

I created this  make command together with @orangecms . Someone might find it useful to run the compiled sources on the Lichee. You need to have a usable "xfel" binary in your path for that :)

### DIFF
--- a/src/mainboard/sunxi/nezha/Makefile
+++ b/src/mainboard/sunxi/nezha/Makefile
@@ -12,6 +12,10 @@ XFEL       ?= xfel
 MEMADDR    ?= 0x40000000
 RELEASE    ?= --release
 
+MODE       ?= release
+BOOT0      ?= $(OREBOOT)/target/$(TARGET)/$(MODE)/oreboot-nezha-bt0.bin
+
+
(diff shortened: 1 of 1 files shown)


src/mainboard/sunxi/nezha/Makefile
```diff
@@ -12,6 +12,10 @@ XFEL       ?= xfel
 MEMADDR    ?= 0x40000000
 RELEASE    ?= --release
 
+MODE       ?= release
```
orangecms, line 15: this should be in sync with `RELEASE` (see above)

How do you feel about putting this into xtask instead? Currently it's a mix already, and gotten messy.

m0veax: Thanks for the feedback. I'll have a look into xtask and will try to implement that next week

rminnich: xtask at first looks hard, but you will be happy if you go for it.

orangecms:
We are currently reworking the build system, with lots of changes pending to already include this here.

Closing thusly.

PR #714, opened 2023-12-22
fixup: visionfive2: use Makefile `PORT` argument

rmsyn:
Uses the supplied `PORT` Makefile variable as an argument to `vf2-loader`.


rmsyn:
The `QuickFormat` check is broken, and failed to download the Rust toolchain. This check should be fixed, or removed the download is fixed.

rmsyn:
The `QuickFormat` check is broken, and failed to download the Rust toolchain. This check should be fixed, or removed until the download is fixed.

orangecms (approved):

PR #713, opened 2023-12-19
feature: VisionFive2 PAC for peripheral access

rmsyn:
Uses the [`jh71xx-pac`](https://github.com/rmsyn/jh71xx-pac) library for peripheral access in the `Visionfive2` mainboard.

PR #715, opened 2023-12-22
feature: visionfive2: use `jh71xx-hal` for peripheral access

rmsyn:
Uses the [`jh71xx-hal`](https://github.com/rmsyn/jh71xx-hal) crate for peripheral access, mainly the `UART0` peripheral.

The `jh71xx-hal` is currently a work-in-progress. Most notably, the `interrupt` code around `PLIC` configuration is apparently broken. The GPIO code also requires a big rewrite to match code from other `embedded-hal` implementations.

Depends on the changes in https://github.com/oreboot/oreboot/pull/713

### DIFF
--- a/src/lib/log/src/lib.rs
+++ b/src/lib/log/src/lib.rs
@@ -103,12 +103,7 @@ impl embedded_hal_nb::serial::Error for Error {
     }
 }
 
-#[cfg(not(feature = "jh71xx"))]
 type SerialLogger = dyn Serial<Error = Error>;
-#[cfg(feature = "jh71xx")]
-type SerialLogger = dyn Serial<Error = jh71xx_hal::uart::Error>;
-#[cfg(feature = "jh71xx")]
-impl<UART: jh71xx_hal::uart::Serial> Serial for jh71xx_hal::uart::Uart<UART> {}
 
 #[cfg(feature = "mutex")]
 static LOGGER: spin::Mutex<Option<&'static mut SerialLogger>> = spin::Mutex::new(None);
@@ -125,7 +120,6 @@ impl fmt::Write for SerialLogger {
             }
             block!(self.write(byte)).ok();
         }
-        #[cfg(not(feature = "jh71xx"))]
         block!(self.flush()).ok();
         Ok(())
     }
--- a/src/lib/log/src/lib.rs
+++ b/src/lib/log/src/lib.rs
@@ -121,12 +121,12 @@ impl fmt::Write for SerialLogger {
         for &byte in s.as_bytes() {
             // Inject a carriage return before a newline
             if byte == b'\n' {
-                block!(self.write(b'\r')).unwrap();
+                block!(self.write(b'\r')).ok();
             }
-            block!(self.write(byte)).unwrap();
+            block!(self.write(byte)).ok();
         }
         #[cfg(not(feature = "jh71xx"))]
-        block!(self.flush()).unwrap();
+        block!(self.flush()).ok();
         Ok(())
     }
 }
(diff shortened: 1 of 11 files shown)


rmsyn:
There are some formatting fixes, and a minor build fix needed. Other than that, ready to review.

src/lib/log/src/lib.rs
```diff
@@ -94,16 +94,21 @@ macro_rules! println {
...
 
-impl embedded_hal::serial::Error for Error {
-    fn kind(&self) -> embedded_hal::serial::ErrorKind {
+impl embedded_hal_nb::serial::Error for Error {
+    fn kind(&self) -> embedded_hal_nb::serial::ErrorKind {
         self.kind
     }
 }
 
+#[cfg(not(feature = "jh71xx"))]
```
orangecms, line 106: Is there a chance that we could avoid specific implementations leaking into this library?

rmsyn: We could, but it would require each implementation to have a HAL that implements the `embedded_hal_nb::serial` traits.

The reason I had to rework the trait here is because `jh71xx-hal` already implements these traits. The other `mainboard` implementations do not. I can try to rework things here again, and see if I can avoid conditional compilation. I agree that would be preferable.

rmsyn: We could, but it would require each implementation to have a HAL that implements the `embedded_hal_nb::serial` traits.

The reason I had to rework the trait here is because `jh71xx-hal` already implements these traits. The other `mainboard` implementations do not. I can try to rework things here again, and see if I can avoid conditional compilation.

Implementations should be moved into each `mainboard` SoC implementation. There would still need to be conditional compilation based on which `mainboard` is being compiled, but could use generic type arguments to hide implementation details.

It's also because you define the type alias `SerialLogger` with the concrete `log::Error` type. Then you store a trait object as a mutable static, which prevents supplying a generic argument if we swapped the concrete `log::Error` for a generic type argument.

The design pattern I've seen in most HALs is that an `Error` type is defined per-module implementing each HAL trait.

As a shim placeholder until each `mainboard` SoC has a HAL crate, the `Serial` and `embedded_hal_nb::serial` traits should be implemented for each `mainboard` SoC UART peripherals. Then, we could do single conditional compilation based on the implementation from each module.

src/soc/src/serial.rs
```diff
@@ -0,0 +1,26 @@
...
+impl fmt::Write for dyn Serial<Error = Error> {
+    fn write_str(&mut self, s: &str) -> Result<(), fmt::Error> {
+        for &byte in s.as_bytes() {
+            // Inject a carriage return before a newline
+            if byte == b'\n' {
+                block!(self.write(b'\r')).ok();
+            }
+            block!(self.write(byte)).ok();
+        }
+        #[cfg(not(feature = "jh71xx"))]
```
rmsyn, line 22: Missed removing this when I moved the implementation. Will remove with the next set of edits.

rmsyn:
Moving the `Serial` trait to the `soc` library will require a lot of rework.

Moving the trait causes cyclic dependencies, so it's going to take me more time to think about the required architecture changes.

src/lib/log/src/lib.rs
rmsyn: Alright, I added a wrapper type in `visionfive2/bt0/src/uart` so we can impl the `log::Serial` trait there.

orangecms:
Awesome! 🥳 I'll try this out. looking good :)

orangecms (approved):
Working just fine! :partying_face:

PR #716, opened 2024-01-09
VisionFive 2 DRAM config fix

orangecms:
It only took me 8 months to figure out this one tiny detail. Mind the comment in the Makefile. Who would have thought...


archanox (approved):

orangecms:
actual docs on this:
https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags

will set this to draft and update; and add something to the docs on Cargo.toml referencing the cfg/rustflags options for build-time config

rmsyn:
This doesn't actually fix the problem, the runtime checks still fail. If you add an extra print statement in `bt0/dram.rs` you can see none of the values are set at runtime: <https://github.com/rmsyn/oreboot/blob/vf2-dram-cfg-fix/src/mainboard/starfive/visionfive2/bt0/src/dram.rs#L12-L17>:

```
DRAM: 8G: false 4G: false 2G: false
```

I also added features for setting the DRAM size, instead of the rustc `cfg` flag, since it seems a bit cleaner. Either way, the value needs to be set in the `features` or `config` flags set in `xtask`, not in the `Makefiles`.

With the changes in my feature branch:

```
DRAM: 8G: true 4G: false 2G: false
```

Currently, the feature is hardcoded, which is obviously the wrong thing, but I wanted a MVP to show proof-of-concept. If we decide to stick with the `--config <target>.rustflags="--cfg DRAM_SIZE=\"<SIZE>\""`, then that would need to be added to `xtask`.

rmsyn (changes requested):
This doesn't actually fix the problem, the runtime checks still fail. If you add an extra print statement in `bt0/dram.rs` you can see none of the values are set at runtime: <https://github.com/rmsyn/oreboot/blob/vf2-dram-cfg-fix/src/mainboard/starfive/visionfive2/bt0/src/dram.rs#L12-L17>:

```
DRAM: 8G: false 4G: false 2G: false
```

I also added features for setting the DRAM size, instead of the rustc `cfg` flag, since it seems a bit cleaner. Either way, the value needs to be set in the `features` or `config` flags set in `xtask`, not in the `Makefiles`.

With the changes in my feature branch:

```
DRAM: 8G: true 4G: false 2G: false
```

Currently, the feature is hardcoded, which is obviously the wrong thing, but I wanted a MVP to show proof-of-concept. If we decide to stick with the `--config <target>.rustflags="--cfg DRAM_SIZE=\"<SIZE>\""`, then that would need to be added to `xtask`.

orangecms:
I have run this code myself;

here is output from `make DRAM_SIZE=8G run`, for example:
```
oreboot 🦀
boot mode: UART
RISC-V arch 8000000000000007
RISC-V core vendor: SiFive (0x0489)
RISC-V implementation: 21G1.02.00 / llama.02.00-general (0x04210427)
RISC-V hart ID 0
apb0 80000000
apb0 80000000
DRAM: 4G: false 2G: false
8G shiny DRAM
[DRAM] init start
[DRAM] set clk to OSC div2
[DRAM] set PLL frequency
PLL1: 00b02603 55e00000 00c7a601
PLL1: 042ba603 41e00000 00c7a60c
```

The `cfg` stuff is evaluated at build time; `features` is for crate options

orangecms:
I have pushed another commit that checks for a valid size option at build time.

Try the following:
- `make run` default to 4G; will work fine
- `make run DRAM_SIZE=2G` will change to 2G, working fine
- `make run DRAM_SIZE=1G` will error because it is not supported

orangecms:
I have added a URL in the Makefile pointing to the upstream docs.

rmsyn:
> I have run this code myself;
> 
> here is output from `make DRAM_SIZE=8G run`, for example:
> 
> ```
> oreboot 🦀
> boot mode: UART
> RISC-V arch 8000000000000007
> RISC-V core vendor: SiFive (0x0489)
> RISC-V implementation: 21G1.02.00 / llama.02.00-general (0x04210427)
> RISC-V hart ID 0
> apb0 80000000
> apb0 80000000
> DRAM: 4G: false 2G: false
> 8G shiny DRAM
> [DRAM] init start
> [DRAM] set clk to OSC div2
> [DRAM] set PLL frequency
> PLL1: 00b02603 55e00000 00c7a601
> PLL1: 042ba603 41e00000 00c7a60c
> ```
> 
> The `cfg` stuff is evaluated at build time; `features` is for crate options

Note that you are not printing the `8G` cfg value:

```
> DRAM: 4G: false 2G: false
```

I ran your code as published, and added a single extra argument to the print statement to show the `8G` `DRAM_SIZE` value. It evaluates to false. Yes, the code compiles, and runs on the board. However, any code blocked behind the `cfg!(dram_size = "8G")` calls is not being run.

> I have added a URL in the Makefile pointing to the upstream docs.

Yes, I read that documentation, and I'm telling you that those changes need to be made in `xtask`. I spent hours trying to get the make files to work, and the only thing that switched on the valid runtime checks is making the changes in `xtask`.

orangecms:
> Note that you are not printing the `8G` cfg value:
> 
> ```
> > DRAM: 4G: false 2G: false
> ```

I am printing this here though (see above):
```
8G shiny DRAM
```

and I have added a commit that prints it more generically.


> I ran your code as published, and added a single extra argument to the print statement to show the `8G` `DRAM_SIZE` value. It evaluates to false. Yes, the code compiles, and runs on the board. However, any code blocked behind the `cfg!(dram_size = "8G")` calls is not being run.

Why did it work for me (see above)? With the last commit, if not passing a supported DRAM size, you would now get an error. Please try that last commit. Note that I force-pushed to amend the `Makefile` change with the second URL.

> Yes, I read that documentation, and I'm telling you that those changes need to be made in `xtask`. I spent hours trying to get the make files to work, and the only thing that switched on the valid runtime checks is making the changes in `xtask`.

Ah... hang on, I was using the `Makefile` in `bt0` directly. Yes, this is finally to be done in `xtask`.
I have other patches for that in the making; they will rework the `xtask` setup.

For a preview: https://github.com/orangecms/oreboot/tree/vf2-xtask/xtask/src

rmsyn:
> Ah... hang on, I was using the Makefile in bt0 directly. Yes, this is finally to be done in xtask.
I have other patches for that in the making; they will rework the xtask setup.

Alright, I'll review after those changes are in, because the compilation step fails when running `make DRAM_SIZE="8G" run` from `mainboard/starfive/visionfive2` directory. It looks like the config options aren't being passed to the `bt0` stage.

orangecms:
okay so the xtask rework prerequisite is in #717

rmsyn:
> okay so the xtask rework prerequisite is in #717

I did an initial review of the code there, and `cherry-pick`ed the changes here on top: https://github.com/rmsyn/oreboot/tree/xtask-shared-utils-vf2

The code still fails to compile with: `make DRAM_SIZE="8G" run` (from `src/mainboard/starfive/visionfive2`)

Are you still working on the fixes in this PR?

orangecms:
> > okay so the xtask rework prerequisite is in #717
> 
> I did an initial review of the code there, and `cherry-pick`ed the changes here on top: https://github.com/rmsyn/oreboot/tree/xtask-shared-utils-vf2
> 
> The code still fails to compile with: `make DRAM_SIZE="8G" run` (from `src/mainboard/starfive/visionfive2`)
> 
> Are you still working on the fixes in this PR?

Yes, it is another small change only, passing the dram_size from xtask (hardcoded in a first step).
I will file that as the next PR. And maybe I will have it at the point then where `--dram_size` becomes an argument already. :)

orangecms:
here is a checkpoint commit - will split this up properly later, but it works nicely now =)

```
oreboot 🦀
boot mode: SPI
RISC-V arch 8000000000000007
RISC-V core vendor: SiFive (0x0489)
RISC-V implementation: 21G1.02.00 / llama.02.00-general (0x04210427)
RISC-V hart ID 0
apb0 80000000
apb0 80000000
DRAM size: 8G
[DRAM] init start
[DRAM] set clk to OSC div2
[DRAM] set PLL frequency
PLL1: 00b02603 55e00000 00c7a601
PLL1: 042ba603 41e00000 00c7a60c
[DRAM] set clk to PLL1 div2
[DRAM] asserts
[DRAM] train
[DRAM] util
[DRAM] start
[DRAM] set clk to OSC div2
[DRAM] boot
[DRAM] OMC init
[DRAM] OMC init train
[DRAM] OMC init PHY
[DRAM] OMC init done
[DRAM] init done
Copy payload... ⏳
"area@0" / "description" Some("header")
"area@0" / "compatible" Some("none")
"area@0" / "size", Some(1024)
"area@1" / "description" Some("bt0")
"area@1" / "compatible" Some("ore-bt0")
"area@1" / "size", Some(65536)
"area@1" / "file" Some("starfive-visionfive2-bt0.bin")
"area@2" / "description" Some("Fixed DTFS")
"area@2" / "compatible" Some("ore-dtfs")
"area@2" / "size", Some(4096)
"area@2" / "file" Some("starfive-visionfive2-board.dtb")
"area@3" / "description" Some("main")
"area@3" / "compatible" Some("ore-main")
"area@3" / "size", Some(65536)
"area@3" / "file" Some("starfive-visionfive2-main.bin")
............ done.
lzss compressed Linux
dump 32 bytes @21400000
1515151515151509150b0b1010100d0d0d0d0d0d0d0d0d0d0d0d0d0d0210190c
dump 32 bytes @21400020
0c0b151515080415151515151515151515151515151515151515151515151515
dump 32 bytes @21400040
1515151515151515151515151515151515151515151515151515151515151515
dump 32 bytes @21400060
1515151515151515151515151515151515151515151515151515151515151515
dump 32 bytes @21400080
1515151515151515151515151515151515151515151515151515151515151515
dump 32 bytes @214000a0
1515151515151515151515151515151515151515151515151515151515151515
dump 32 bytes @214000c0
151515151515151515151515151515151515151515151515151515151515100d
dump 32 bytes @214000e0
0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d0d
release non-boot harts =====

Jump to payload... with dtb 0x8010000
```

rmsyn (approved):

orangecms:
happy to tell that this works splendid :sparkles:

xtask/src/main.rs
```diff
@@ -63,6 +63,26 @@ impl FromStr for Memory {
...
+    TwoG,
+    FourG,
+    EightG,
+}
+
+impl FromStr for DramSize {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
```
rmsyn, line 77:
```suggestion
        match s.as_str() {
```

Since you are checking upper and lowercase strings below, you can skip the conversion (and save a heap allocation).

xtask/src/main.rs
```diff
@@ -63,6 +63,26 @@ impl FromStr for Memory {
...
+            "2g" | "2G" => Ok(Self::TwoG),
```
rmsyn, line 78:
```suggestion
            "2g" | "2G" => Ok(Self::TwoG),
```

If you decide to keep the `to_lowercase` conversion above, you don't need to match the uppercase versions (since they'll never be present).

rmsyn (commented):
Couple of nits, but nothing blocking. Changes LGTM

rminnich (approved):

xtask/src/main.rs
orangecms: ah yes - will keep that in mind for subsequent refinement, thank you!

PR #717, opened 2024-01-11
xtask shared utils

orangecms:
This is a major rework that vastly reduces the boilerplate for our build tasks.

### DIFF
--- a/xtask/src/starfive/visionfive2.rs
+++ b/xtask/src/starfive/visionfive2.rs
@@ -1,138 +1,55 @@
-use crate::dist_dir;
-use crate::layout_flash;
-use crate::project_root;
-use crate::{Commands, Env};
-use fdt;
-use log::{error, info, trace};
-use std::fs::File;
-use std::io::{self, Seek, SeekFrom};
-use std::process::{self, Command, Stdio};
+use crate::util::{
+    compile_board_dt, dist_dir, find_binutils_prefix_or_fail, get_cargo_cmd_in, objcopy,
+    project_root,
+};
+use crate::{layout_flash, Cli, Commands, Env};
+// use fdt;
+use log::{error, info, trace, warn};
+use std::{
+    fs::{self, File},
+    io::{self, Seek, SeekFrom},
+    path::Path,
+    process,
+};
 
-use std::{env, fs, path::Path};
 extern crate layoutflash;
 use layoutflash::areas::{create_areas, Area};
 
-// It is not clear what CRC starfive had in mind, they wrote their own ...
-fn crc32_reverse(x: u32) -> u32 {
-    let mut x = x;
-    x = ((x & 0x55555555) << 1) | ((x >> 1) & 0x55555555);
-    x = ((x & 0x33333333) << 2) | ((x >> 2) & 0x33333333);
-    x = ((x & 0x0F0F0F0F) << 4) | ((x >> 4) & 0x0F0F0F0F);
-    (x << 24) | ((x & 0xFF00) << 8) | ((x >> 8) & 0xFF00) | (x >> 24)
-}
+use super::visionfive2_hdr::spl_create_hdr;
 
-fn crc32(iv: u32, sv: u32, data: Vec<u8>) -> u32 {
-    let mut crc = iv;
-    for v in data {
-        let mut byte = crc32_reverse(v.into());
-        for _x in 0..8 {
-            crc = if ((crc ^ byte) & 0x80000000u32) != 0 {
-                (crc << 1) ^ sv
-            } else {
-                crc << 1
-            };
-            byte = byte << 1;
-        }
-    }
+const HEADER_SIZE: usize = 0x400;
 
-    crc
-}
+const ARCH: &str = "riscv64";
+const TARGET: &str = "riscv64imac-unknown-none-elf";
 
-fn crc32_final(iv: u32) -> u32 {
-    crc32_reverse(iv ^ !0u32)
-}
+const BT0_BIN: &str = "starfive-visionfive2-bt0.bin";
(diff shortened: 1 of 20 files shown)


xtask/src/util.rs
```diff
@@ -0,0 +1,114 @@
...
+    process::{self, Command, Stdio},
+};
+
+/// These utilities help find and run external commands.
+/// Those are mostly toolchain components and vendor specific tools.
+
+/// This gets you the `cargo` command in a specific directory.
+/// Use it to build a stage of a mainboard, which is a board's subdirectory.
+pub fn get_cargo_cmd_in(env: &Env, root: PathBuf, dir: &str, command: &str) -> Command {
+    let cargo = std::env::var("CARGO").unwrap_or_else(|_| "cargo".to_string());
```
rmsyn, line 14: Nit:

```suggestion
    let cargo = std::env::var("CARGO").unwrap_or("cargo".to_string());
```

xtask/src/sunxi/nezha.rs
```diff
@@ -62,153 +63,100 @@ pub(crate) fn execute_command(args: &crate::Cli, features: Vec<String>) {
...
     if !features.is_empty() {
         let command_line_features = features.join(",");
-        trace!("append command line features: {}", command_line_features);
+        trace!("append command line features: {command_line_features}");
         command.arg("--no-default-features");
         command.args(["--features", &command_line_features]);
     } else {
         trace!("no command line features appended");
     }
+    trace!("run D1 bt0 build command: {:?}", command);
```
rmsyn, line 94: Nit:

```suggestion
    trace!("run D1 bt0 build command: {command:?}");
```

xtask/src/starfive/visionfive2.rs
```diff
@@ -1,266 +1,126 @@
...
-            xtask_add_bt0_header(&args.env);
-            // main stage
-            xtask_build_jh7100_flash_main(&args.env);
-            xtask_binary_jh7100_flash_main(binutils_prefix, &args.env);
-            xtask_concat_flash_binaries(&args.env);
+            // Get binutils first so we can fail early
+            let binutils_prefix = &find_binutils_prefix_or_fail(ARCH);
+            // Build the stages - should we parallelize this?
+            xtask_build_jh7110_main(&args.env);
+            xtask_build_jh7110_bt0(&args.env, &features);
```
rmsyn, line 47: This may not matter, but is there a reason for switching the build order of `main` and `bt0`?

xtask/src/starfive/visionfive2_hdr.rs
```diff
@@ -0,0 +1,119 @@
...
+        let rc32 = crc::Crc::<u32>::new(&crc::CRC_32_ISCSI);
+        let mut digest = rc32.digest();
+        digest.update(&dat);
+        let crcout = digest.finalize();
+    }
+    */
+    let v = crc32(CRC_IV, CRC_SV, dat.clone());
+    let fv = crc32_final(v);
+
+    let mut hdr = vec![];
```
rmsyn, line 70: We can save potential re-allocations here with:

```suggestion
    let mut hdr = Vec::with_capacity(METADATA_LEN + dat.len());
```

Where `METADATA_LEN` is the fixed-length of the known metadata (`HEADER_OFFSET`, `BACKUP`, `VERSION`, etc.).

rmsyn:
Other than some small changes, and a few minor nits, the code LGTM.

Boards still compile.

I'll need time to do a more in-depth review, but left comments after an initial pass.

xtask/src/starfive/visionfive2.rs
orangecms: Yes, it should not matter here. It is a relic of iteration of approaches. I'll swap it back.

xtask/src/starfive/visionfive2_hdr.rs
orangecms: Sounds good!

orangecms:
> Other than some small changes, and a few minor nits, the code LGTM.

Awesome, thank you so much - I will push the respective fixups. :)

> I'll need time to do a more in-depth review, but left comments after an initial pass.

Alright!

xtask/src/starfive/visionfive2_hdr.rs
orangecms: let's postpone that part for later, and keep this step here down to just moving out the header cration to its own file

xtask/src/starfive/visionfive2.rs
orangecms: done

xtask/src/sunxi/nezha.rs
orangecms: done

xtask/src/util.rs
orangecms: done

orangecms:
I have found a few issues with the VF2 build; I'm looking into that. The current stuff is a bit weird.
We ought to build via layoutflash, and the final function for that needs some extras for adding the header.
I had set up another task for doing that, concatenating the binaries, etc..

Set to draft for now...

xtask/src/starfive/visionfive2_hdr.rs
rmsyn: > let's postpone that part for later, and keep this step here down to just moving out the header cration to its own file

Why? The majority of the header size is known, the only dynamic size is the input data. Adding `Vec::with_capacity` has no downsides in this case (we are not over-allocating), and saves potentially costly re-allocation(s).

orangecms: I would just like to keep the scope down here. Optimizations can follow later - the intent of this PR is to factor things out to make the build system more comprehensible. :-)

rminnich (approved):

PR #719, opened 2024-01-13
visionfive2: use `jh71xx-pac` for `clk_cpu_root`

orangecms:
taken from #713


orangecms:
dropped in favor of taking the whole #713

PR #720, opened 2024-01-13
VisionFive 2 rework loading the main stage via DTFS

orangecms:


orangecms:
this comes after #716

src/mainboard/starfive/visionfive2/bt0/src/main.rs
```diff
@@ -80,10 +69,11 @@ pub unsafe extern "C" fn start() -> ! {
         "csrw   mtvec, zero",
         // 1. suspend non-boot hart
         // hart 0 is the S7 monitor core; 1-4 are U7 cores
-        "li     a1, 0",
+        "li     a1, 1",
         "csrr   a0, mhartid",
         "bne    a0, a1, .nonboothart",
         // 2. prepare stack
+        // FIXME: each hart needs its own stack
```
rmsyn, line 76: Since we are loading a payload, isn't the payload responsible for creating its runtime stacks?

Also, we are only running in the supervisor core with a `riscv64imac` target. All the other cores are running under the `riscv64gc` target (a superset of `riscv64imac`). I'm not sure about the safety of doing any setup that crosses that boundary. Thoughts?

src/mainboard/starfive/visionfive2/bt0/src/main.rs
```diff
@@ -225,140 +277,85 @@ fn main() {
...
-        let base = 0x0800_0000 + 32 * 1024;
-        let dram = DRAM_BASE;
-        for b in (0..0x100).step_by(4) {
-            write32(dram + b, read32(base + b));
-            if b % 0x4_0000 == 0 {
-                print!(".");
-            }
+        (SRAM0_BASE, SRAM0_SIZE) // occupied space
+    };
+    let slice = unsafe { core::slice::from_raw_parts(transmute(base), size) };
```
rmsyn, line 313: Maybe add a `SAFETY` comment here since we are loading executable memory from FLASH/RAM?

src/mainboard/starfive/visionfive2/bt0/src/main.rs
```diff
@@ -225,140 +277,85 @@ fn main() {
...
-        DRAM_BASE + 0x0020_0000
-    } else {
-        DRAM_BASE
-    };
-    // println!("Payload @{load_addr:08x}");
-
-    let hart_id = mhartid::read();
-    // U-Boot proper
+fn exec_payload(addr: usize) {
     unsafe {
```
rmsyn, line 354: Similar to above, maybe add a `SAFETY` comment about `transmute`ing executable memory into an `EntryPoint`?

Since we're not doing a signed/secure boot setup, we're assuming the memory is valid, but I think that should at least be mentioned.

src/mainboard/starfive/visionfive2/bt0/src/main.rs
rmsyn, line 313: Maybe add a `SAFETY` comment here since we are loading executable memory from FLASH/RAM?

src/mainboard/starfive/visionfive2/bt0/src/main.rs
rmsyn, line 354: Similar to above, maybe add a `SAFETY` comment about `transmute`ing executable memory into an `EntryPoint`?

Since we're not doing a signed/secure boot setup, we're assuming the memory is valid, but I think that should at least be mentioned.

src/mainboard/starfive/visionfive2/bt0/src/main.rs
rmsyn, line 76: Since we are loading a payload, isn't the payload responsible for creating its runtime stacks?

Also, we are only running in the supervisor core with a `riscv64imac` target. All the other cores are running under the `riscv64gc` target (a superset of `riscv64imac`). I'm not sure about the safety of doing any setup that crosses that boundary. Thoughts?

rmsyn (commented):
Few questions, and suggested `SAFETY` comments.

rminnich (approved):
seems good enough to me, but you might want to see about the other comments.

src/mainboard/starfive/visionfive2/bt0/src/main.rs
orangecms: To clarify: _supervisor_ is a mode in RISC-V; a core is called a _hart_ - and the SoC starts each hart in machine mode.
I think you mean _monitor core_, which I think is what they call hart 0 in the manual and the [data sheet](https://doc-en.rvspace.org/JH7110/PDF/JH7110_Datasheet.pdf), an S7 (RV64IMAC) that does not have the supervisor mode / has no MMU, since it's an [embedded core](https://www.sifive.com/blog/the-sifive-20g1-update-for-7-series-core-ip) (not that well documented...).
That is why it cannot run the same Linux kernel as the others and why we need to switch to running on harts 1-4, while one of them needs to be chosen to do the initialization that we do in oreboot and has control over the peripherals, e.g. for serial output. That is why this PR includes the commit to switch to running on hart 1.
Via the optional SBI that we want to offer you would drop into S-mode. That is what #694 is about.

The compile target means that we use only that set (subset) of instructions; that has no impact on the cores, they only need to support the respective set of used instructions.

src/mainboard/starfive/visionfive2/bt0/src/main.rs
orangecms: The _secure boot_ topic is a very good point.

If we reverse engineer the previous vendor tool that apparently had some signing capabilities and we figure out how to burn keys into the SoC, then we could offer secure boot; though that is somewhat outside the current scope. :D It would be cool to have though and I'm sure we can do it!

We do jump to code under our own control here, since you would build oreboot fully, and so it is no more or less secure than the rest of the code.

The point on safety - same with the above - is indeed something to take care of. Let's add those notes.

orangecms: Re secure boot; see https://github.com/starfive-tech/Tools/issues/1 where I reversed said tool for the simpler header generation that we also have in oreboot

src/mainboard/starfive/visionfive2/bt0/src/main.rs
rmsyn: > The compile target means that we use only that set (subset) of instructions; that has no impact on the cores, they only need to support the respective set of used instructions.

Right, I guess it would only matter going in the opposite direction (GC to IMAC), since the former includes floating point instructions, and the latter doesn't.

Anyway, thanks for the detailed response.

src/mainboard/starfive/visionfive2/bt0/src/main.rs
rmsyn: > It would be cool to have though and I'm sure we can do it!

I think we could just choose an unused offset in the QSPI flash, write a public key there, and load it into memory on boot. Not perfect, since an attacker with write access to the QSPI could rewrite with their own key, but they could also do whatever they want at that point.

Then in the `bt0`/SPL (or some earlier stage), we load the key + signature, and verify whatever signed pieces of firmware and/or kernel we want.

PR #723, opened 2024-01-21
feature: vf2: use `jh71xx-pac` for DDR configuration

rmsyn:
Uses the `jh71xx-pac` library for DDR configuration in the VisionFive2 mainboard.

### DIFF
--- a/src/mainboard/starfive/visionfive2/bt0/src/ddr_start.rs
+++ b/src/mainboard/starfive/visionfive2/bt0/src/ddr_start.rs
@@ -313,30 +313,25 @@ pub unsafe fn start() {
     let phy = &*pac::DMC_PHY::ptr();
 
     START_CFG0.iter().for_each(|cfg| {
-        phy.ac_base(cfg.reg_nr as usize).modify(|r, w| {
-            w.bits((r.bits() & cfg.mask) | cfg.value)
-        });
+        phy.ac_base(cfg.reg_nr as usize)
+            .modify(|r, w| w.bits((r.bits() & cfg.mask) | cfg.value));
     });
     START_CFG1.iter().for_each(|cfg| {
-        phy.base(cfg.reg_nr as usize).modify(|r, w| {
-            w.bits((r.bits() & cfg.mask) | cfg.value)
-        });
+        phy.base(cfg.reg_nr as usize)
+            .modify(|r, w| w.bits((r.bits() & cfg.mask) | cfg.value));
     });
     // NOTE: Commented out in VF1 code
     if cfg!(dram_size = "2G") {
-        phy.base(11).modify(|r, w| {
-            w.bits((r.bits() & 0xffff_fff0) | 0x0000_0005)
-        });
+        phy.base(11)
+            .modify(|r, w| w.bits((r.bits() & 0xffff_fff0) | 0x0000_0005));
     }
     START_CFG2.iter().for_each(|cfg| {
-        phy.base(cfg.reg_nr as usize).modify(|r, w| {
-            w.bits((r.bits() & cfg.mask) | cfg.value)
-        });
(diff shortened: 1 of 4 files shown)


orangecms:
The purpose of the params was to make a reusable library that could also apply to other SoCs using the same controller. I Noticed that the JH7100 init code looked very much the same, just with some different values and different base addresses. How do you feel about that?

orangecms:
> > How do you feel about that?
> 
> I don't have a VF1 to test on, but I can try to use the JH7110 code on the VisionFive1 board.

I have a VF1, and the DRAM init for it is a standalone C implementation that I took and translated as well back then; that's the code in oreboot. It has many comments documenting the registers.

> One thing that might be an issue is if the DDR memory controller has different base addresses in the JH7100. Seeing as there also isn't a public reference manual for the JH7100 anymore, I don't really have any reference for comparing.

There was never a manual, only a short datasheet:

https://starfivetech.com/uploads/JH7100%20Datasheet.pdf

rmsyn:
> I have a VF1, and the DRAM init for it is a standalone C implementation that I took and translated as well back then; that's the code in oreboot. It has many comments documenting the registers.

Right, what I'm saying is those registers are different enough that a separate module in `jh71xx-pac` will be needed for the VF1 peripherals. Since I don't have a datasheet/reference manual to go on, I will have to go through the source code available for the board, and create a SVD. That's a LOT of work.

orangecms:
Could you just run the formatter again? :grimacing:

orangecms:
> > I have a VF1, and the DRAM init for it is a standalone C implementation that I took and translated as well back then; that's the code in oreboot. It has many comments documenting the registers.
> 
> Right, what I'm saying is those registers are different enough that a separate module in `jh71xx-pac` will be needed for the VF1 peripherals. Since I don't have a datasheet/reference manual to go on, I will have to go through the source code available for the board, and create a SVD. That's a LOT of work.

yea... for reference, here is the code for the JH7100 that I translated back then - you will see the close similarity:
https://github.com/starfive-tech/JH7100_ddrinit/blob/starfive/ddrc_cfg/lpddr4_1600_cl28_bl16/orbit_boot_training.c

rmsyn:
> yea... for reference, here is the code for the JH7100 that I translated back then - you will see the close similarity

Sure, and it will take adding those register definitions into an SVD to generate code using `svd2rust`. That process would need to be repeated for each peripheral that is not the same as what's on the JH7110.

You're welcome to take a look at what I've added to the [`cmsis-svd-generator`](https://github.com/rmsyn/cmsis-svd-generator/tree/dev), and fill out similar python scripts for JH7100 peripherals.

I'm also working on [`svd-generator`](https://codeberg.org/weathered-steel/svd-generator), which is a Rust version of `cmsis-svd-generator`. Most of the plumbing for adding new peripherals is already there, so if you wanted to make a feature branch for JH7100 peripherals, I would welcome a merge request!

orangecms:
Tested to work fine - I suggest revisiting the potential to definine parameters and factoring this out for some other point in time, maybe when JH8xxx SoCs are out and turn out to be similar enough again or something.

orangecms:
For upcoming PRs, please do us the favor to squash the fixups into the respective corresponding commits that are to be fixed up, so that browing the git history becomes easier. Thanks!

rmsyn:
> For upcoming PRs, please do us the favor to squash the fixups into the respective corresponding commits that are to be fixed up, so that browing the git history becomes easier.

Interesting... I was planning to squash/fixup the fixup commit before this PR was merged, but you rebased my work on the latest `main`, cross-signed them with your key, and merged the commit.

Please refrain from doing this to mine and other contributors' work. It muddles the git history on the authorship when running a `git blame` on the affected files.

You also failed to request squashing the fixup commit until right before merging this PR, even though the commit was present for over a month.

PR #731, opened 2024-03-01
Update to embedded-hal 1.0 throughout the project

felipebalbi:


felipebalbi:
Fails due to missing #728

orangecms (approved):
Nice, thank you!

this tarpaulin thing ... let's kick it out, serves no value at this point.

felipebalbi:
@orangecms even after removing `tarpaulin`, we still need to Bump rust version because of some `clap` dependency:

```
Caused by:
  package `clap_complete v4.5.1` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.73.0
```

We still need to merge #728 before any of the other PRs will work.

felipebalbi:
Yeah, anything under the `clap` umbrella has MSRV listed to rust 1.74, see https://github.com/clap-rs/clap/blob/f65d421607ba16c3175ffe76a20820f123b6c4cb/Cargo.toml#L17

orangecms (approved):

orangecms:
Gotcha, thank you so much for clearing all of this out!

PR #732, opened 2024-03-05
Add aarch64 emulation

felipebalbi:
First we teach `xtask` how to compile RiscV emulation and proceed with re-adding `qemu-aarch64` target.


felipebalbi:
For now I'm really just looking for confirmation that I'm in the right path. Do we want to revisit `qemu-aarch64` or do we start it all from scratch? Note that I removed all the code pointing at `oreboot::drivers` and the aarch64 emulation target is basically a stub of what it should be doing.

I can see life with qemu + gdb, but there's no output of any kind. Any suggestions on how to add proper PL011 support? There are some crates, but they depend on cortex-m.

@orangecms FYI

felipebalbi:
Failing due to lack of #728

orangecms:
Looks perfectly fine to me :)

For QEMU, the approach from what I am thinking of would be to really just have the main stage and test the higher-level features - decompression, running some test payload, and seeing that our setup works fine (whatever it'd be on Arm, for RISC-V it'd be the SBI, etc). Does that make sense to you as well?

If you don't mind, I'd love to take that first commit here already. Would you move that over into another PR? :)

orangecms:
For the PL011, you can use the register block definition from here:
https://github.com/apgoetz/pl011_qemu/blob/master/src/lib.rs

just copy that over, essentially, and implement the embedded HAL trait for it and the oreboot log library:
https://github.com/oreboot/oreboot/blob/main/src/lib/log/src/lib.rs

PR #733, opened 2024-03-06
Remove tarpaulin dependency

felipebalbi:


orangecms:
we'll need to remove this from CI as well :sweat_smile: so far it had served no value and only created noise... x)

PR #735, opened 2024-03-07
xtask/starfive/visionfive2: rework header creation

orangecms:
The struct is based on what the original vendor tool does. No clue yet how the secure boot flow or setup really works. Reference: https://github.com/starfive-tech/Tools/issues/8

Add a Default impl.

Use CRC_32_ISO_HDLC from the crc crate as figured out by https://github.com/jonirrings/vf2-header :)


kanakshilledar (approved):

kanakshilledar:
LGTM @orangecms

PR #736, opened 2024-03-07
Teach xtask how to build for qemu-riscv

felipebalbi:


felipebalbi:
Hmm, there's a bit more work required to ensure CI continues to run after adding qemu-riscv to `xtask`.

orangecms (approved):

PR #745, opened 2024-05-31
arch/riscv: comment on register saving/restoring

orangecms:


luojia65 (approved):
Looks good, it simplifies internal function names and comments, and enhances readability for developers.

PR #747, opened 2024-06-01
arch/riscv: rework SBI debug prints

orangecms:


orangecms:
This turned out to be very useful during development.

luojia65 (approved):
In the future, we can use `#[cfg(feature = "...")]` to replace the current work on debugging.

PR #749, opened 2024-06-21
nezha: Fix first jump

ArthurHeymans:
The KEEP linker directive does not prohibit the rust compiler to garbage collect code, unless #[used] is used. Therefore remove MainStageHead as it's effectively not there.

The first jump instruction did assume MainStageHead was there and therefore skipped the first 2 instructions of _start. This is now fixed.

This changes the first instruction to be in assembly rather than inline assembly as the rust compiler seems to like to pad the instruction with unimp.

This means that disabling interrupts and enabling "theadisaee" and "maee" is now correctly done.

### DIFF
--- a/src/mainboard/sunxi/nezha/bt0/src/main.rs
+++ b/src/mainboard/sunxi/nezha/bt0/src/main.rs
@@ -42,26 +42,15 @@ const STACK_SIZE: usize = 1 * 1024; // 1KiB
 #[link_section = ".bss.uninit"]
 static mut BT0_STACK: [u8; STACK_SIZE] = [0; STACK_SIZE];
 
-/// Jump over head data to executable code.
-///
-/// # Safety
-///
-/// Naked function.
-#[naked]
-#[link_section = ".head.text"]
-#[export_name = "_head_jump"]
-pub unsafe extern "C" fn head_jump() {
-    asm!(
-        ".option push",
-        ".option rvc",
-        "c.j    0x68", // 0x60: eGON.BT0 header; 0x08: FlashHead
-        ".option pop",
-        // sym _start,
-        options(noreturn)
-    )
-}
-
-// todo: option(noreturn) generates an extra `unimp` insn
+// Use global assembly for a 4 byte jump instruction to _start.
+// The reason is that rust adds an extra `unimp` insn after the jump if inline assembly is used.
+// This messes up the location of the eGON header
+core::arch::global_asm!(
+    ".section .head.text, \"ax\"",
+    ".global _head_jump",
+    "_head_jump:",
+    "    j start"
+);
 
 // eGON.BT0 header. This header is identified by D1 ROM code
 // to copy BT0 stage bootloader into SRAM memory.
@@ -92,7 +81,7 @@ const DTF_SIZE: usize = 0x1_0000; // 64K
 const LIN_SIZE: usize = 0x00fc_0000;
 const DTB_SIZE: usize = 0x0001_0000;
 
-// clobber used by KEEP(*(.head.egon)) in link script
+#[used]
 #[link_section = ".head.egon"]
 pub static EGON_HEAD: EgonHead = EgonHead {
     magic: *b"eGON.BT0",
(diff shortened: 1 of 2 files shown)


ArthurHeymans:
Not yet tested though

rminnich (approved):
I'm always ready to get rid of inline assembly when I can, so I tend to like this. I hope it's ok.

orangecms:
The current code works. We just use a compressed jump. What is being fixed here exactly?

The header as needed by the mask ROM for execution requires the first bytes to be a jump to the actual code, and after it follows metadata. Its size must be exactly what we currently have.

src/mainboard/sunxi/nezha/bt0/src/main.rs
```diff
@@ -42,26 +42,10 @@ const STACK_SIZE: usize = 1 * 1024; // 1KiB
...
-        ".option pop",
-        // sym _start,
-        options(noreturn)
-    )
-}
-
-// todo: option(noreturn) generates an extra `unimp` insn
+// Use assembly file for a 4 byte jump instruction to _start.
+// The reason is that rust adds an extra `unimp` insn after the jump.
+// This messes up the location of the eGON header
```
orangecms, line 47: I cannot confirm this.
The current code works and has been working for years now.
How did you experience the issue described here?

orangecms: The current instruction is compressed, on that note, i.e. 2 bytes, and we need 4, so the other 2 bytes are necessary.

ArthurHeymans: So the current code jumps a bit too far. "// 0x60: eGON.BT0 header; 0x08: FlashHead" Is not correct. Flashhead is not part of the image so the jump should be to 0x60, not 0x68. I don't think disabling interrupts (might be reset default) and enabling the Thead specific extensions are crucial, which is why things worked. Using a 4 byte jump, just like u-boot (4 first bytes are exactly the same) seemed convenient.

I tried tried inline assembly jump to a symbol, rather than a raw number, but then the rust compiler pads the jump too much with another 4 bytes of nothing. This is what my comment meant.

orangecms: Oooh... gotcha; so can't we just change the 0x68 to 0x60 then?
I'd prefer the inlining so that it's all in one place, but of course it should also be correct.

orangecms:
```
Disassembly of section .head:

0000000000020000 <_head_jump>:
   20000:       a0a5                    j       20068 <_start+0x8>
        ...

0000000000020004 <_ZN17oreboot_nezha_bt09EGON_HEAD17h20872eadf5e41c7cE>:
   20004:       4765                    li      a4,25
   20006:       422e4e4f                .4byte  0x422e4e4f
   2000a:       3054                    .2byte  0x3054
   2000c:       6c39                    lui     s8,0xe
   2000e:       5f0a                    lw      t5,160(sp)
        ...

Disassembly of section .text:

0000000000020060 <_start>:
   20060:       30401073                csrw    mie,zero
   20064:       00600337                lui     t1,0x600
   20068:       7c032073                csrs    0x7c0,t1
   2006c:       000303b7                lui     t2,0x30
   20070:       23cd                    addiw   t2,t2,19 # 30013 <ebss+0xbf68>
   20072:       7c239073                csrw    0x7c2,t2
```

this is what currently happens, yea

orangecms: and what you're really saying is that MainStageHead was never effectively used - and it is probably not needed?

orangecms: https://github.com/u-boot/u-boot/blob/fe2ce09a0753634543c32cafe85eb87a625f76ca/include/sunxi_image.h#L35

looking at this, it appears that most of the data (all 0 in our current code) isn't used anyway... funny

orangecms:
How do you feel about #750?

ArthurHeymans:
> How do you feel about #750?

It's roughly the same. I think this one has the slight edge as you can remove one extra line from the linker script (KEEP(*(.text.entry))) and it require less explaining what the inline asm does.

src/mainboard/sunxi/nezha/bt0/src/main.rs
ArthurHeymans: > Oooh... gotcha; so can't we just change the 0x68 to 0x60 then? I'd prefer the inlining so that it's all in one place, but of course it should also be correct.

As the file indeed contains just 1 instruction, I moved it to using strings in the main.rs file.

orangecms (approved):
Sweet!

orangecms:
the linter is a bit unhappy about the extra newline :grimacing: 

we have a global `make format` target, that should fix it

orangecms (approved):

orangecms:
woop woop - thank you!

PR #750, opened 2024-06-22
sunxi/nezha/bt0: simplify eGON header

orangecms:
based on ideas from #749


orangecms:
I'll eventually rebase onto #749

ArthurHeymans (approved):
I suppose those are some u-boot leftovers

PR #753, opened 2024-06-23
sunxi: Boot from SPI flash at 100MHz using fast read

ArthurHeymans:
While JEDEC_READ is typically capped at a lower frequency. Fast read can clock much higher at the expense of a extra 8 bits to send. More than doubling the frequency at the expense of 20% higher data overhead is worth it.

Note that some older SPI flash might not like this higher frequency though...


orangecms:
This hasn't worked on first try; maybe a lower frequency will do.

PR #755, opened 2024-08-31
README: Use HTTPS instead of the anonymous git:// protocol

cyrozap:
GitHub disabled cloning via the anonymous git:// protocol [in 2022](https://github.blog/changelog/2022-03-15-removed-unencrypted-git-protocol-and-certain-ssh-keys/), so documentation should suggest users clone via HTTPS instead.


orangecms (approved):

orangecms:
Thank you!

PR #759, opened 2024-12-08
Docs: dilemma of drivers and RISC-V updates

orangecms:


README.md
```diff
@@ -132,6 +132,25 @@ To get a general understanding of how oreboot and firmware in general works,
...
 describes how firmware is stored and boots up on a platform / SoC.
 
+## Dilemma of drivers
+
+Note that oreboot does not aim to turn into its own operating system.
+Accordingly, we intend to keep the amount and functionality of drivers low.
+However, by design of SoCs, we do have to implement some things over time:
+
+- SPI flash, SD cards and eMMC storage are rather simple to load from, but
+  require mechanisms to flash over time
```
neuschaefer, line 142: I had to read this twice.

```diff
-  require mechanisms to flash over time
+  require mechanisms to write over time
```

... seems a bit clearer

neuschaefer (changes requested):
minor comment

README.md
orangecms: yea good point... flash/write/burn/program - so many words! Should we roll a :game_die: or use all four? :D

orangecms: mechanisms is maybe also not that great, rather _mechanisms/tools_...

orangecms: How about now?

neuschaefer: > yea good point... flash/write/burn/program - so many words! Should we roll a 🎲 or use all four? :D

- **burn**: feels kind of anachronistic by now (remember those old EEPROMs where you had to apply a double-digit programming voltage? that was before my time)
- **program**: somewhat confusing because it has two meanings: the human act of constructing a program, and the mechanical act of writing code/data to a memory medium
- **flash**: easily confused with a noun (the name of a specific, albeit ubiquitous technology
- **write**: I do think this is a good generic term, and can be used alone

orangecms: Good points! I'm sticking with _write_, reworded it a bit to make it even more clear

neuschaefer (approved):

kanakshilledar (approved):
LGTM

PR #760, opened 2024-12-15
refactor: return &'static Path when returning PathBuf is unnecessary

Integral-Tech:
- When returning a path, prefer using `&'static Path` instead of `PathBuf` when heap allocation is unnecessary.


Integral-Tech:
Any progress on this pull request?

orangecms (approved):
Thank you!

PR #766, opened 2025-02-11
Cargo.toml: use * to include all boards and lib crates

orangecms:
This makes it simpler to new boards, stages, and libraries.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,18 +5,8 @@ members = [
   "src/console",
   "src/cpu",
   "src/soc",
-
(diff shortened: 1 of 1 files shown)


rminnich (approved):

rminnich:
I can't remember why we stopped doing that, but it's nice to see it back.

orangecms:
> I can't remember why we stopped doing that, but it's nice to see it back.

It took some experimentation to see that it makes sense this way.
We had multiple iterations with various approaches of wildcards.

PR #768, opened 2025-02-14
canaan/canmv-k230/main: clean up SBI platform timer

orangecms:


src/mainboard/canaan/canmv-k230/main/src/sbi_platform.rs
```diff
@@ -97,35 +89,12 @@ impl rustsbi::Timer for Timer {
...
-        if DEBUG {
-            let v = read64(mtime_cmp);
-            println!("[SBI] current mtime cmp: {v:016x}");
-        }
 
         // Reenable the interrupt
-        if DEBUG {
-            println!("[SBI] enable timer interrupt");
-        }
         unsafe { mie::set_mtimer() }
```
rminnich, line 98: just wondering, can's the mid::set_mtimer do the unsafe inside and not require this?

rminnich (approved):

src/mainboard/canaan/canmv-k230/main/src/sbi_platform.rs
orangecms: The csr macros in the riscv crate define them all to be unsafe, so no;
https://github.com/rust-embedded/riscv/blob/1dac3619aca697491f5cf2dc275d1e729d2de9f9/riscv/src/register/macros.rs#L398

In general, a CSR may not be defined, or may error depending on execution mode, which cannot be guaranteed, only by the programmer, so we have to wrap the CSR access in `unsafe {}`.

PR #769, opened 2025-02-18
add more utils

orangecms:


src/lib/util/src/mem.rs
```diff
@@ -0,0 +1,102 @@
...
+    let print_step = size / step_size / 64;
+
+    println!("Memory test: {size}bytes @ 0x{base:016x}");
+
+    println!("Memory test: write patterns...");
+    for (i, a) in (base..limit).step_by(step_size).enumerate() {
+        if i % print_step == 0 {
+            print!(".");
+        }
+        #[allow(clippy::identity_op)]
```
rminnich, line 38: I have not tried this, but could you do
let i = I as u32
and then just use I in the next four lines?

src/lib/util/src/mmio.rs
```diff
@@ -0,0 +1,40 @@
...
+    unsafe { read_volatile(address as *mut u32) }
+}
+
+pub fn write32(address: usize, value: u32) {
+    unsafe { write_volatile(address as *mut u32, value) }
+}
+
+// NOTE: Some hardware does not like direct u64 MMIO access.
+// So read low and high word (little endian) one at a time, then merge them.
+pub fn read64(reg: usize) -> u64 {
```
rminnich, line 29: This is going to burn you at some point, the problem with this is that some hardware will discard high order bits when you read32 the low 32 bits ... it can really suck that way. I wonder if you want to make this explicit, like
read64_hl
or something. 
i.e. make it really clear in the name that this does not do a real read64?

src/lib/util/src/mmio.rs
```diff
@@ -0,0 +1,40 @@
...
+    let l = read32(reg) as u64;
+    let h = read32(reg + 4) as u64;
+    (h << 32) | l
+}
+
+// NOTE: Some hardware does not like direct u64 MMIO access.
```
rminnich, line 35: same. I had a severe driver bug once because the 64-bit register would consider a 32-bit read/write as a 64-bit read/write and it did not end well.

src/lib/util/src/mem.rs
rminnich: e.g. 

fn main() {
    let i: u32 = 0;
    println!("Hello, world!{i}");
    let i = i as i32;
    println!("Hello, world!{i}");
}

src/lib/util/src/mem.rs
```diff
@@ -0,0 +1,102 @@
...
+use crate::mmio::{read32, write32};
+
+// NOTE: Print an empty line _before_ instead of doing println!(), such that
+// after progress dots, there is always a line break before the error message.
+fn check32(addr: usize, val: u32) -> Result<(), ()> {
+    let v = read32(addr);
+    if v != val {
+        print!("\nError @ {addr:08x}: expected {val:08x}, got {v:08x}");
+        Err(())
+    } else {
```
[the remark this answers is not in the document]
orangecms: Either you do an explicit return before the `Err(())`, and then omit the `else`, or you use the return-last-stament convention, then you need the else.
```rs
if foo {
  return Err(());
}
Ok(())
```
vs
```rs
if foo {
  Err(())
} else {
  Ok(())
}
```

No big diff, either works.

src/lib/util/src/mem.rs
```diff
@@ -0,0 +1,102 @@
...
+        Ok(())
+    }
+}
+
+const TEST_PATTERN0: u32 = 0x2233_ccee;
+const TEST_PATTERN1: u32 = 0x5577_aadd;
+const TEST_PATTERN2: u32 = 0x1144_bbff;
+const TEST_PATTERN3: u32 = 0x6688_9900;
+
+pub fn test(base: usize, size: usize) {
```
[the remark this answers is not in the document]
orangecms: With the current simple `fn test`, you can `use util::mem;` and later `mem::test`.
You could always go with an alias and `use util::mem::test as mem_test;` if you like.
I personaly prefer the `mem::test` to avoid `mem::mem_test`.

src/lib/util/src/mem.rs
orangecms: We could... an extra line/assignment vs another inline cast - I'm sure the compiler would optimize it either way.
I personally find the few little inlines okay here.

src/lib/util/src/mmio.rs
orangecms: Alright, I'll do that.

orangecms: I will go with `read64le`.

rminnich (approved):

src/lib/util/src/mem.rs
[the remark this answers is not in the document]
orangecms: It would just be a statement, but not return. For the implicit return, you need to omit the `;` and have no further statement. Otherwise, it has to be explicit.

Issue #772, opened 2025-02-27
(de)compression


orangecms:
So far, we are using the `lzss` crate for (de)compression to fit a kernel Image on a SPI flash.

This has its pros and cons, and we can do better, with drawbacks.

For a general intro to approaches to using (de)compressors, see:
https://timclicks.dev/article/how-to-use-compression-algorithms-in-rust

### Time

lzss takes a lot of time to compress, decompression is mostly fine.

Note that compression happens on the host machine that builds the firmware image, so any cost there is mostly a one-time cost and only "bad" for the development experience when it is slow.

### Size

The compressed size with lzss is okayish; given a 17 MB Linux Image with few drivers that already includes an xz-compressed u-root initramfs with many commands (nothing to expect in that part), we end up with 11 MB.

In comparison, the same Image is compressed much faster and down to 8 MB using zlib-rs and the highest compression level (9).

### To alloc or not...

However, zlib-rs requires an allocator if we want to use it for decompression: https://github.com/trifectatechfoundation/zlib-rs/issues/309
So does https://docs.rs/zopfli/latest/zopfli/.

OTOH, `lzss` works without a heap allocator, and so does https://github.com/Frommi/miniz_oxide with `default-features = false`.

### More options

Is there a better (de)compression library that we can use without an allocator?

Should we implement one in order to get better compression?

It looks like almost all libraries need an allocator: https://blog.logrocket.com/rust-compression-libraries/

See also 

- https://github.com/KillingSpark/zstd-rs/
- https://github.com/dropbox/rust-brotli
- https://github.com/BurntSushi/rust-snappy
- https://github.com/pseitz/lz4_flex

dancrossnyc:
I was surprised to see that you said `miniz_oxide` requires an allocator, but I remembered that I provide one in `phbl` so perhaps I never noticed. Regardless, it appears that you can turn it off in `Cargo.toml`.  From https://crates.io/crates/miniz_oxide,

>Using the library with default-features = false removes the dependency on alloc and collection crates, making it suitable for systems without an allocator. Running without allocation reduces crate functionality:
>
> * The deflate module is removed completely
> * Some inflate functions which return a Vec are removed

[...]

PR #780, opened 2025-03-06
sunxi/nezha fixes + rework

orangecms:
This is in preparation of #772; the current only SoC we use our compression library for is the D1, so that's to be switched, and as it turns out, we had some breakage - and could already benefit  from our newly introduced XuanTie library.


orangecms:
Note: I have already tested `miniz_oxide` locally - it works just fine for the D1, and the next PR will implement the switch, being quite trivial.

luojia65 (approved):

PR #781, opened 2025-03-07
switch to miniz_oxide for compression

### LINKED ISSUE #772: (de)compression

orangecms:
This required adjusting the sunxi/nezha main stage due to the API change, but also because we need a bigger stack now.
While at it, rework the constants and prints a bit.
With the experience gained now, more rework will follow.

This resolves #772 for a start. We have yet to come to a clear design for building and flashing.

### DIFF
--- a/src/lib/compression/src/lib.rs
+++ b/src/lib/compression/src/lib.rs
@@ -1,45 +1,45 @@
 #![no_std]
 
-use core::{ptr::read_volatile, slice};
+use core::slice;
 use log::println;
 
-// offset_bits aka EI, usually 10..13
-// length_bits aka EJ, usually 4..5
-// buf_fill_byte (often 0x20, space) - we use 0x00, the zero byte
-// decompress_buf_size, always 1 << EI
-// compress_buf_size, always 2 << EI
-const EI: usize = 12;
-const DECOMP_BUF_SIZE: usize = 1 << EI;
-const COMP_BUF_SIZE: usize = 2 << EI;
-pub type OreLzss = lzss::Lzss<EI, 4, 0x00, DECOMP_BUF_SIZE, COMP_BUF_SIZE>;
+use miniz_oxide::inflate as moi;
+use miniz_oxide::inflate::core as moc;
+
+// This flag means inflate should try parsing a zlib header.
+// We provide zlib compressed payloads including the header by convetion.
+const FLAGS: u32 = moc::inflate_flags::TINFL_FLAG_PARSE_ZLIB_HEADER;
 
 /// # Safety
 ///
 /// The caller needs to ensure correctness of source and target addresses.
-/// The first four bytes at source address must hold the compressed size.
 /// The payload size is the maximum expected size after decompression.
 /// Ensure that it fits, i.e., `payload_size` bytes after target address
 /// are unused. We allocate a buffer of that size; exceeding it will error.
-pub unsafe fn decompress(source_addr: usize, target_addr: usize, payload_size: usize) {
-    // NOTE: This must be u32 because it is 4 bytes, so cast later.
-    let compressed_size = unsafe { read_volatile(source_addr as *const u32) };
-    let in_ptr = (source_addr + 4) as *const u8;
+pub unsafe fn decompress(
+    source_addr: usize,
+    target_addr: usize,
+    compressed_size: usize,
+    payload_size: usize,
+) {
+    let in_ptr = source_addr as *const u8;
     let out_ptr = target_addr as *mut u8;
-    let input = unsafe { slice::from_raw_parts(in_ptr, compressed_size as usize) };
+    let input = unsafe { slice::from_raw_parts(in_ptr, compressed_size) };
     let output = unsafe { slice::from_raw_parts_mut(out_ptr, payload_size) };
 
     println!(
         "Decompress {compressed_size} bytes from {input:p} to {output:p}, reserved {payload_size} bytes",
     );
 
-    let reader = lzss::SliceReader::new(input);
-    let writer = lzss::SliceWriter::new(output);
-    match OreLzss::decompress_stack(reader, writer) {
-        Ok(r) => {
-            println!("Success, decompressed {r} bytes :)");
+    let mut decompressor = moc::DecompressorOxide::new();
+    let (status, _, out_bytes) = moc::decompress(&mut decompressor, &input, output, 0, FLAGS);
+
+    match status {
+        moi::TINFLStatus::Done => {
+            println!("Success, decompressed {out_bytes} bytes :)");
         }
-        Err(e) => {
-            panic!("Decompression error {e}");
+        _ => {
+            panic!("Decompression error {status:?}");
         }
     }
 }
--- a/src/mainboard/sunxi/nezha/main/src/main.rs
+++ b/src/mainboard/sunxi/nezha/main/src/main.rs
@@ -21,53 +21,47 @@ use spin;
 
 mod sbi_platform;
 
-const MEM: usize = 0x4000_0000;
+const RAM_BASE: usize = 0x4000_0000;
 
 // see ../fixed-dtfs.dts
-// const PAYLOAD_OFFSET: usize = 0x2_0000;
-const PAYLOAD_SIZE: usize = 0x20_0000; // 2 MB
-const PAYLOAD_ADDR: usize = MEM + 0x20_0000;
-
-// compressed image
-const LINUXBOOT_TMP_OFFSET: usize = 0x0400_0000;
-const LINUXBOOT_TMP_ADDR: usize = MEM + LINUXBOOT_TMP_OFFSET;
-
 // target location for decompressed image
-const LINUXBOOT_OFFSET: usize = 0x0020_0000;
-const LINUXBOOT_ADDR: usize = MEM + LINUXBOOT_OFFSET;
-const LINUXBOOT_SIZE: usize = 0x0200_0000;
-// DTB_OFFSET should be >=LINUXBOOT_OFFSET+LINUXBOOT_SIZE and match bt0
-// TODO: Should we just copy it to a higher address before decompressing Linux?
-const DTB_OFFSET: usize = LINUXBOOT_OFFSET + LINUXBOOT_SIZE;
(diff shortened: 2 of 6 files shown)


src/lib/compression/src/lib.rs
```diff
@@ -1,45 +1,45 @@
...
 /// The caller needs to ensure correctness of source and target addresses.
-/// The first four bytes at source address must hold the compressed size.
 /// The payload size is the maximum expected size after decompression.
 /// Ensure that it fits, i.e., `payload_size` bytes after target address
 /// are unused. We allocate a buffer of that size; exceeding it will error.
-pub unsafe fn decompress(source_addr: usize, target_addr: usize, payload_size: usize) {
-    // NOTE: This must be u32 because it is 4 bytes, so cast later.
-    let compressed_size = unsafe { read_volatile(source_addr as *const u32) };
-    let in_ptr = (source_addr + 4) as *const u8;
+pub unsafe fn decompress(
```
dancrossnyc, line 19: Why make this unsafe? I presume it's because the source and destination arguments are passed as `usize`? Why not just have it take a slice, at least for the source? I would define this as something like,

    pub fn decompress(source: &[u8], dst: &mut [u8]) { ...

src/lib/compression/src/lib.rs
```diff
@@ -1,45 +1,45 @@
...
+    source_addr: usize,
+    target_addr: usize,
+    compressed_size: usize,
+    payload_size: usize,
+) {
+    let in_ptr = source_addr as *const u8;
     let out_ptr = target_addr as *mut u8;
-    let input = unsafe { slice::from_raw_parts(in_ptr, compressed_size as usize) };
+    let input = unsafe { slice::from_raw_parts(in_ptr, compressed_size) };
     let output = unsafe { slice::from_raw_parts_mut(out_ptr, payload_size) };
```
dancrossnyc, line 28: This does not seem save to me; taking a slice of uninit memory is UB. If you memset it to zero first, that's ok. But I'd say it's even better for this function to just take a mutable slice of bytes to expand into.

src/mainboard/sunxi/nezha/main/src/main.rs
```diff
@@ -21,53 +21,47 @@ use spin;
...
-const LINUXBOOT_ADDR: usize = MEM + LINUXBOOT_OFFSET;
-const LINUXBOOT_SIZE: usize = 0x0200_0000;
-// DTB_OFFSET should be >=LINUXBOOT_OFFSET+LINUXBOOT_SIZE and match bt0
-// TODO: Should we just copy it to a higher address before decompressing Linux?
-const DTB_OFFSET: usize = LINUXBOOT_OFFSET + LINUXBOOT_SIZE;
-const DTB_ADDR: usize = MEM + DTB_OFFSET;
+const COMPRESSED_ADDR: usize = RAM_BASE + 0x0400_0000;
+const COMPRESSED_SIZE: usize = 0x00fe_0000;
 
 fn decompress_lb() {
```
dancrossnyc, line 38: Note that this doesn't just decompress; it does a lot of checking and validation, too.

src/mainboard/sunxi/nezha/main/src/main.rs
```diff
@@ -21,53 +21,47 @@ use spin;
...
     let dtb = unsafe { read_volatile(DTB_ADDR as *const u32) };
     if dtb != 0xedfe0dd0 {
-        panic!("DTB looks wrong: {:08x}\n", dtb);
+        panic!("DTB looks wrong: {dtb:08x}");
     } else {
-        print!("DTB looks fine, yay!\n");
+        println!("DTB looks fine, yay!");
     }
 
     unsafe {
```
dancrossnyc, line 47: So what I'd do here is something like,

        let compressed_addr = core::ptr::with_exposed_provenance::<u8>(COMPRESSED_ADDR);
        let compressed = unsafe { slice::from_raw_parts(compressed_addr, COMPRESSED_SIZE);
        let payload_addr = core::ptr::with_exposed_provenance_mut::<u8>(PAYLOAD_ADDR);
        let payload = unsafe {
            core::ptr::write_bytes(payload_addr, 0, PAYLOAD_SIZE);
            slice::from_raw_parts_mut(payload_addr, PAYLOAD_SIZE);
        };
        decompress(compressed, payload);

src/mainboard/sunxi/nezha/main/src/main.rs
```diff
@@ -21,53 +21,47 @@ use spin;
...
+    let a = PAYLOAD_ADDR + 0x30;
     let r = unsafe { read_volatile(a as *mut u32) };
     if r == u32::from_le_bytes(*b"RISC") {
-        print!("Payload looks like Linux Image, yay!\n");
+        println!("Payload looks like Linux Image, yay!");
     } else {
-        panic!("Payload does not look like Linux Image: {:x}\n", r);
+        panic!("Payload does not look like Linux Image: {r:x}");
     }
     // Recheck on DTB, kernel should not run into it
```
dancrossnyc, line 59: What does this mean? It seems superfluous; if the concern is that the DTB might overlap the payload that you're decompressing above, then a) that's a violation of Rust's memory safety rules, since you're constructing a mutable slice in `decompress` right now that would alias an existing structure, and b) I would test for that specifically, and not as a side-effect of decompression.  Here, you should be able to leverage the language and its guarantees to avoid duplicating the check.

src/mainboard/sunxi/nezha/main/src/main.rs
```diff
@@ -344,9 +336,9 @@ extern "C" fn main() -> usize {
         }
         println!("You are NOT MY SUPERVISOR!");
         unsafe {
-            decompress(LINUXBOOT_TMP_ADDR, PAYLOAD_ADDR, PAYLOAD_SIZE);
+            decompress(COMPRESSED_ADDR, PAYLOAD_ADDR, COMPRESSED_SIZE, PAYLOAD_SIZE);
```
dancrossnyc, line 339: See the notes above vis slices and so forth; they apply here as well.

dancrossnyc (commented):
A number of comments here. More generally, this change is conflated with a bunch of semi-related cleanups (`print!("whatever\n");` to `println!("whatever");` and similar). I'd have separated those into their own change; it was a tad hard to see what was going on.

orangecms:
> A number of comments here. More generally, this change is conflated with a bunch of semi-related cleanups (`print!("whatever\n");` to `println!("whatever");` and similar). I'd have separated those into their own change; it was a tad hard to see what was going on.

Thank you and sorry for the mess! This is the first port that we had in oreboot so far that allows us to boot a kernel now; more rework is planned with the experience we now have. I'll reply to the respective comments.

src/lib/compression/src/lib.rs
orangecms: Yes, the idea here was that the caller wouldn't need to take care of creating the slices and doing the necessary pointer/address conversions. That is effectively what this library here does, besides a few prints. Otherwise, we wouldn't need it, and a lot of code would be duplicated across multiple ports. Hence `unsafe`, and the safety note, having the caller fully responsible of providing suitable addresses and sizes.

src/mainboard/sunxi/nezha/main/src/main.rs
orangecms: Yea, this is no longer necessary; this is from earlier iterations and can be removed now.

src/mainboard/sunxi/nezha/main/src/main.rs
orangecms: Interesting, [`with_exposed_provenance()`](https://doc.rust-lang.org/stable/core/ptr/fn.with_exposed_provenance.html) is new to me - introduced in Rust 1.84.0 per the docs. I will need to read up on it in more detail. Thank you!

src/lib/compression/src/lib.rs
dancrossnyc: It's not that it may mess with the allocator per se, but rather that the behavior is undefined because it violates the language's rules.

luojia65 (approved):

src/lib/compression/src/lib.rs
orangecms: Yea I totally agree, especially,

> because all four arguments are the same type, there's no good way to prevent messing them up

is a concern I also have. I think the best approach is then to really drop the compression library here altogether and just have the few extra lines in every oreboot port.

To provide more context:
The currently hardcoded addresses were very ad-hoc to get this whole thing working for a start. The eventual goal is indeed to predefine the memory regions at build time somehow. That is where the compression takes place anyway. I would like to get the responsibility to ensure that both sides fit together and the memory regions are safe to use into the build system (xtask), since that is where the image is stitched anyway, so I think it makes most sense to have the memory region assertions there.

PR #782, opened 2025-03-08
sunxi/nezha: do the decompression directly without our library

orangecms:
Following a discussion on the compression library design, it is clear that a) the interface is hard to get right, and b) safety guarantees need to be on the caller side, so we may as well have the slice creation there and then won't win much with the library.

see #781 for reference


src/mainboard/sunxi/nezha/main/src/main.rs
```diff
@@ -312,16 +316,38 @@ extern "C" fn main() -> usize {
...
+        let dtb = unsafe { read_volatile(DTB_ADDR as *const u32) };
+        if dtb != DT_MAGIC {
+            panic!("DTB looks wrong: {dtb:08x}");
+        } else {
+            println!("DTB looks fine, yay!");
+        }
+
+        // check for kernel to be okay
+        let a = PAYLOAD_ADDR + 0x30;
+        let r = unsafe { read_volatile(a as *const u32) };
```
dancrossnyc, line 337: I don't think you need `read_volatile` here. Regular `read` should suffice.

rminnich:
can we continue with this one or does it get dropped.

PR #788, opened 2025-05-17
README rework

orangecms:
remove the old cruft, add guidance on adding targets

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,37 +253,6 @@ make mainboards
 make -j mainboards
 ```
 
-## QEMU
-
-```
-# Install QEMU for your target platform, e.g. x86
-sudo apt install qemu-system-x86
-
-# Build release build and start with QEMU
-cd src/mainboard/emulation/qemu-q35 && make run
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -314,10 +280,6 @@ oreboot seeks to support:
 
 ## Ground Rules
 
-- `Makefile`s must be simple. Use `xtask` instead for control flow, e.g., adding
-  headers or checksums to the binaries, sitchting images, etc..
```
kanakshilledar, line 318: I think we should keep this point, even though it is being mentioned in the `xtask/README.md`.
Rest all looks good!

orangecms: Yea my goal here is to reduce the main README and push those details to where they apply. Those rules have in part become implicit over time with the code as it is now. And let's face it - there is about noone else on the project now. I want to encourage participation and provide positive examples rather than enforcing strict rules.

kanakshilledar (approved):
LGTM

PR #791, opened 2025-09-13
canmv-k230/bt0: use common crates, cleanup

orangecms:


src/mainboard/canaan/canmv-k230/bt0/src/main.rs
```diff
@@ -188,15 +188,22 @@ const DO_MEMTEST: bool = false;
 const LOADER: usize = mem_map::MASK_ROM_BASE;
 
 // TODO: move to SoC lib crate
-const CPU0_X: usize = mem_map::HDI_BASE + 0x0020;
-const CPU1_X: usize = mem_map::HDI_BASE + 0x0030;
-// We don't know why, but this enables the mtimer clock.
+const STC2_CFG: usize = mem_map::STC_BASE + 0x0020;
```
UtsavAgarwalADI, line 191: nit: Do we want to separate these changes into a separate commit?

UtsavAgarwalADI:
They look good overall, definitely cleaner. Since we are adding some new information though in addition to cleanup, I am wondering if we need to split it into 2 commits?

PR #803, opened 2025-10-01
starfive/visionfive2/bt0: configure cfg lints

orangecms:
Cargo would now warn on features it doesn't know.
Instead of just declaring what features can take what values, they have to be listed in the `lints.rust` config section. See also: https://blog.rust-lang.org/2024/05/06/check-cfg/


kanakshilledar (approved):

kanakshilledar (approved):
LGTM

PR #807, opened 2025-10-03
xtask: define and use common helper for platform directory

orangecms:
reworked to apply immediately


orangecms:
Further iterations will determine the platform directory directly, parse the ELF and bin file names from the crate manifests (`Cargo.toml`), rework the `compile_board_dt()` helper, and more.

rminnich (approved):

PR #809, opened 2025-10-04
Makefile: use cargo wrapper for clippy

orangecms:


m0veax (approved):
Good addition, appreciate

PR #810, opened 2025-10-04
xtask: introduce helpers for cargo manifest and config

orangecms:
As a first step, use them for sunxi/nezha (Allwinner D1). This gets us rid of hardcoding a second copy of the binary names. We do, however, name the _stages_, so define structs for them. Other helpers from the utils will be migrated to use the new Bin struct as their arguments, so we do not need to, e.g., pass target, ELF name and binary name to `objcopy()`, just a `&Bin`.


xtask/src/sunxi/nezha.rs
m0veax, line 1: great, no hardcoded artifacts anymore!

m0veax (approved):

PR #814, opened 2025-10-05
VF1 xtask rework

orangecms:
based on #813


orangecms:
> Greatly reduces the amount of constants needed.
> 
> `BT0_STAGE`, `MAIN_STAGE` constants and the `Stages` struct seem a bit redundant but that might be on me not knowing where else that is used.
> 
> At the very least it's neater than before that way.

Yea, there is going to be another use for them later; might be that that can be reduced yet again, later.

code4brot (approved):

PR #816, opened 2025-10-05
QEMU/RISC-V xtask rework

orangecms:
Move objcopy() into the main stage build function and compile_board_dt() into the DTFS image build function, respectively.

like #815


kanakshilledar (approved):
LGTM

PR #824, opened 2026-02-21
fix make run on starfive mainboards

SpieringsAE:
These artifacts that are to be flashed are not actually put into $(OREBOOT)/target/... but in their respective mainboard directory.


orangecms:
Oops, sorry, this is a half-baked thing I left here.

The recovery loader is mostly meant to run bt0.bin only, except sometimes I added a bit more to it, but a full image (with, say, a Linux kernel) wouldn't really fit. 💾

Eventually, the run command is meant to be integrated in our xtask based build system. I'll see what to do next here. 🧐

For context, I am looking for collaborators to improve and fully implement said framework, because it's currently at a draft/50% level. 🙃

orangecms:
Addendum: the image to be flashed needs a `flash` command, but we're lacking a usable flashing tool for the JH71x0 platforms at this point.

orangecms (approved):

PR #828, opened 2026-04-07
rework stages in layoutflash

orangecms:
Use HashMap to hold the stages. Fill them by conventional names instead of files that are easy to get wrong.

I had this cooking for months alreday, time to get it off the stove. :shallow_pan_of_food:

### DIFF
--- a/src/lib/layoutflash/src/layout.rs
+++ b/src/lib/layoutflash/src/layout.rs
@@ -1,9 +1,15 @@
 use std::prelude::rust_2021::*;
 
-use std::io::{self, Seek, SeekFrom, Write};
-use std::{env, fs, path::Path};
+use std::{
+    collections::HashMap,
+    env, fs,
+    io::{self, Seek, SeekFrom, Write},
+    path::{Path, PathBuf},
+};
 
 use crate::areas::{Area, AREAS_PATH};
+
+const EMPTY_BYTE: u8 = 0xff;
 
 /// Create the areas from the FDT.
 pub fn create_areas<'a>(fdt: &'a fdt::Fdt<'a>) -> Result<Vec<Area<'a>>, String> {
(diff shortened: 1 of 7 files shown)


src/lib/layoutflash/src/layout.rs
```diff
@@ -60,54 +103,40 @@ pub fn layout_flash(dir: &Path, path: &Path, areas: Vec<Area>) -> io::Result<()>
...
 
         println!("<{name}> @ 0x{last_area_end:x}");
-        // First fill with 0xff.
+        // First fill with empty bytes.
+        // TODO: rework to just filling up
         let mut v = Vec::new();
-        v.resize(size, 0xff);
+        v.resize(size, EMPTY_BYTE);
         f.seek(SeekFrom::Start(offset as u64))?;
         f.write_all(&v)?;
```
kanakshilledar, line 111: Why are we initializing the vector with empty bytes instead of directly populating values?

xtask/src/qemu/riscv.rs
```diff
@@ -16,7 +17,7 @@ use crate::util::{
 use crate::{Commands, Env};
 
 // const SRAM0_SIZE = 128 * 1024;
-const SRAM0_SIZE: u64 = 32 * 1024;
+// const SRAM0_SIZE: u64 = 32 * 1024;
```
kanakshilledar: Can we get rid of these comments?
Or just add a note instead?

kanakshilledar (changes requested):

src/lib/layoutflash/src/layout.rs
orangecms: We need to prefill the buffer; for a flash part, "clear" means `0xff`.
Mind that this change here is not concerned with it, it just makes it a const.

orangecms: Also note that there can be gaps anywhere.

xtask/src/qemu/riscv.rs
orangecms: we can drop them now, yea

orangecms: done

src/lib/layoutflash/src/layout.rs
kanakshilledar: Okay

kanakshilledar (approved):
LGTM

orangecms:
Excellent, more to come! 🥳

PR #829, opened 2026-04-25
rework building board -> platform device tree

orangecms:
following up on #828, this is another step toward _platform_ instead of _mainboard_ and convention over per-case definitions


xtask/src/util.rs
```diff
@@ -8,6 +8,7 @@ use std::{
 // These utilities help find and run external commands.
 // Those are mostly toolchain components and vendor specific tools.
 
+// TODO: change into `"src/platform"`.
 pub const PLATFORM_BASE_PATH: &str = "src/mainboard";
```
kanakshilledar, line 12: Is the move from `src/mainboard` to `src/platform` a big refactor and scoped for another PR?

kanakshilledar (changes requested):
LGTM, can be merged after the question is resolved!

xtask/src/util.rs
orangecms: Yes indeed!

kanakshilledar (approved):
LGTM

PR #830, opened 2026-04-25
lib/layoutflash: implement Display for Area

orangecms:
this gets us nicer output

### DIFF
--- a/src/lib/layoutflash/src/areas.rs
+++ b/src/lib/layoutflash/src/areas.rs
@@ -1,3 +1,5 @@
+use core::fmt::{Formatter, Result as FmtResult};
+
 use fdt::{node::FdtNode, Fdt, FdtError};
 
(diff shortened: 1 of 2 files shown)


kanakshilledar (changes requested):
Can you please increase the length of your commit message? Will be easier to track changes later in the future :)

orangecms:
> Can you please increase the length of your commit message? Will be easier to track changes later in the future :)

Not sure I'm following - could you explain a bit more?

kanakshilledar:
The commit message just has the subject saying lib/layoutflash: implement Display for Area, but the body is empty. It would be nicer if you make the commit message longer by either saying "why this change is required?" or "what are you trying to change?"

orangecms:
Ah I see what you mean, done!

kanakshilledar (approved):
LGTM

PR #836, opened 2026-05-11
move adding header to xtask for sunxi platforms

orangecms:
Also factor out reusable parts. They will be shared across SoCs.


orangecms:
needs some adjustments, carved out from a larger change, fixups coming soon

rminnich (approved):

PR #843, opened 2026-05-17
xtask/util: rework finding binutils prefix

orangecms:
This is more idiomatic, so to say.


orangecms:
At some point, we might just get rid of this altogether, since we're using `cargo-binutils`, which gives us all the LLVM binutils with a `rust-` prefix; i.e. `rust-objcopy` etc..

rminnich (approved):

rminnich (approved):

PR #854, opened 2026-05-21
rename Bin->Stage

orangecms:
#853 goes first

### DIFF
--- a/xtask/src/starfive/visionfive1.rs
+++ b/xtask/src/starfive/visionfive1.rs
@@ -1,10 +1,4 @@
-use std::{
-    collections::HashMap,
-    fs::{self, File},
-    io::{self, Seek, SeekFrom},
-    path::PathBuf,
-    process,
-};
+use std::{collections::HashMap, fs, path::PathBuf, process};
 
 use fdt::Fdt;
 use log::{error, info, trace};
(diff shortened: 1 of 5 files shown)


rminnich (dismissed):
stages --> nice, it's a well understood term.

xtask/src/starfive/visionfive1.rs
```diff
@@ -1,41 +1,33 @@
...
 
 use crate::util::{
-    compile_platform_dt, find_binutils_prefix_or_fail, get_bin_for, get_cargo_cmd_in, objcopy,
-    platform_dir, target_bin, target_dir, Bin,
+    compile_platform_dt, find_binutils_prefix_or_fail, get_cargo_cmd_in, get_stage_for, objcopy,
+    platform_dir, target_bin, Stage,
 };
 use crate::{Cli, Commands, Env};
 
 // const SRAM0_SIZE = 128 * 1024;
```
kanakshilledar, line 14: Can we get rid of this comment?

kanakshilledar (dismissed):

xtask/src/starfive/visionfive1.rs
orangecms: That is not the scope here at this point.

I want to get to having the build tool understand the memory ranges, so that it becomes the source of truth.

rminnich (approved):

PR #855, opened 2026-05-21
rename lib/{layoutflash->cast_iron}

orangecms:
Also drop the legacy authors field from Cargo.toml.


rminnich (commented):
I don't get the rationale for this name change at all. layoutflash is very clear to me, cast_iron? I don't get it.

kanakshilledar:
+1 to @rminnich's comment.

orangecms:
For some more context:

The flash layout is only one part of this library, and I personally find it confusing to name things after what they are for (or were at some point).
My intention is to make this general the core library of the build system and accumulate the framework here.

The `Stage` struct is one thing I would like to move here, and I will also take a look at this:
https://github.com/OpenDevicePartnership/embedded-services/tree/main/partition-manager/partition-manager

rminnich:
image
imager
cons
constructor
let's work on words. I'm really sorry, but cast_iron I find hard to type and understand.

